PR #9, opened 2020-03-11
libcloudforensics refactoring and unit testing

giovannt0:
This PR brings the following changes:

- Unit testing for the `GoogleCloudProject` class in `gcp_test.py`.
- Refactor of `gcp.py` to align with the coding guidelines described in the [contributing](https://github.com/google/cloud-forensics-utils/blob/master/docs/contributing.md) doc.
- Typos.

### DIFF
--- a/libcloudforensics/gcp.py
+++ b/libcloudforensics/gcp.py
@@ -58,9 +58,9 @@ def create_service(service_name, api_version):
   try:
     credentials = GoogleCredentials.get_application_default()
   except ApplicationDefaultCredentialsError as error:
-    raise RuntimeError(
-      'Could not get application default credentials: {0!s}\n'
-      'Have you run $ gcloud auth application-default login?'.format(error))
+    error_msg = 'Could not get application default credentials: ' \
+                'Have you run $ gcloud auth application-default login?'
+    raise RuntimeError(error_msg)
 
   service_built = False
   for retry in range(RETRY_MAX):
@@ -78,14 +78,13 @@ def create_service(service_name, api_version):
       break
 
   if not service_built:
-    raise RuntimeError(
-      'Failures building service {0:s} caused by multiple timeouts'.format(
-        service_name))
+    error_msg = 'Failures building service {0:s} caused by multiple timeouts'.format(service_name)
+    raise RuntimeError(error_msg)
 
   return service
 
 
-class GoogleCloudProject(object):
+class GoogleCloudProject():
   """Class representing a Google Cloud Project.
 
   Attributes:
@@ -253,9 +252,8 @@ class GoogleCloudProject(object):
     instances = self.list_instances()
     instance = instances.get(instance_name)
     if not instance:
-      error = 'Instance {0:s} was not found in project {1:s}'.format(
-        instance_name, self.project_id)
-      raise RuntimeError(error)
+      error_msg = 'Instance {0:s} was not found in project {1:s}'.format(instance_name, self.project_id)
+      raise RuntimeError(error_msg)
 
     if not zone:
       zone = instance['zone']
@@ -277,9 +275,8 @@ class GoogleCloudProject(object):
     disks = self.list_disks()
     disk = disks.get(disk_name)
     if not disk:
-      raise RuntimeError(
-        'Disk {0:s} was not found in project {1:s}'.format(
-          disk_name, self.project_id))
+      error_msg = 'Disk {0:s} was not found in project {1:s}'.format(disk_name, self.project_id)
+      raise RuntimeError(error_msg)
 
     if not zone:
       zone = disk['zone']
@@ -319,10 +316,11 @@ class GoogleCloudProject(object):
         project=self.project_id, zone=self.default_zone, body=body).execute()
     except HttpError as exception:
       if exception.resp.status == 409:
-        raise RuntimeError('Disk {0:s} already exists'.format(disk_name))
-      raise RuntimeError(
-        'Unknown error (status: {0:d}) occurred when creating '
-        'disk from snapshot:\n{1!s}'.format(exception.resp.status, exception))
+        error_msg = 'Disk {0:s} already exists'.format(disk_name)
+        raise RuntimeError(error_msg)
+      error_msg = 'Unknown error (status: {0:d}) occurred when creating disk ' \
+                  'from snapshot:\n{1!s}'.format(exception.resp.status, exception)
+      raise RuntimeError(error_msg)
     self.gce_operation(operation, zone=self.default_zone, block=True)
     return GoogleComputeDisk(
       project=self, zone=self.default_zone, name=disk_name)
@@ -492,9 +490,8 @@ class GoogleCloudProject(object):
       ex: {'instance-1': {'zone': 'us-central1-a', 'labels': {'id': '123'}}
     """
     if not isinstance(filter_union, bool):
-      raise RuntimeError((
-                           'filter_union parameter must be of Type boolean'
-                           ' {0:s} is an invalid argument.').format(filter_union))
+      error_msg = 'filter_union parameter must be of Type boolean {0:s} is an invalid argument.'.format(filter_union)
+      raise RuntimeError(error_msg)
 
     resource_dict = dict()
     filter_expression = ''
@@ -578,9 +575,8 @@ class GoogleCloudFunction(GoogleCloudProject):
     try:
       json_args = json.dumps(args)
     except TypeError as e:
-      raise RuntimeError(
-        'Cloud function args [{0:s}] could not be serialized: {1!s}'.format(
-          str(args), e))
+      error_msg = 'Cloud function args [{0:s}] could not be serialized: {1!s}'.format(str(args), e)
+      raise RuntimeError(error_msg)
 
     function_path = 'projects/{0:s}/locations/{1:s}/functions/{2:s}'.format(
       self.project_id, self.region, function_name)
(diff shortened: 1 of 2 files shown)


libcloudforensics/gcp.py
```diff
@@ -494,19 +493,19 @@ def __list_by_label(self, labels_filter, service_object, filter_union):
     """
     if not isinstance(filter_union, bool):
       raise RuntimeError((
-          'filter_union parameter must be of Type boolean'
-          ' {0:s} is an invalid argument.').format(filter_union))
+                           'filter_union parameter must be of Type boolean'
+                           ' {0:s} is an invalid argument.').format(filter_union))
```
tomchop: There seems to be an extra set of `()` around this string.

libcloudforensics/gcp_test.py
```diff
@@ -16,28 +16,343 @@
...
+  def test_format_log_message(self):
+    """Test formatting log message. """
     msg = 'Test message'
-    formatted_msg = self.project.format_log_message(msg)
+    formatted_msg = FAKE_ANALYSIS_PROJECT.format_log_message(msg)
     self.assertIsInstance(formatted_msg, six.string_types)
     self.assertEqual(
-        formatted_msg, u'project:{0:s} {1:s}'.format(
-            self.project.project_id, msg))
+      formatted_msg, u'project:{0:s} {1:s}'.format(
```
tomchop: We're python3 only, so the leading `u''` in strings are not necessary

libcloudforensics/gcp.py
```diff
@@ -673,16 +672,16 @@ def form_operation(self, operation_name):
...
     module = None
     if resource_type not in ['compute#instance', 'compute#snapshot',
                              'compute#disk']:
       raise RuntimeError((
-          'Compute resource Type {0:s} is not one of the defined types in '
-          'libcloudforensics library (Instance, Disk or Snapshot) '
-      ).format(resource_type))
+                           'Compute resource Type {0:s} is not one of the defined types in '
+                           'libcloudforensics library (Instance, Disk or Snapshot) '
+                         ).format(resource_type))
```
tomchop: Consider building this string in a `msg =` variable before for better readability

libcloudforensics/gcp.py
```diff
@@ -718,38 +717,38 @@ def add_labels(self, new_labels_dict, blocking_call=False):
...
     operation = None
     if resource_type not in ['compute#instance', 'compute#snapshot',
                              'compute#disk']:
       raise RuntimeError((
-          'Compute resource Type {0:s} is not one of the defined types in '
-          'libcloudforensics library (Instance, Disk or Snapshot) '
-      ).format(resource_type))
+                           'Compute resource Type {0:s} is not one of the defined types in '
+                           'libcloudforensics library (Instance, Disk or Snapshot) '
+                         ).format(resource_type))
```
tomchop: Consider building this string in a `msg =` variable before for better readability

libcloudforensics/gcp_test.py
```diff
@@ -16,28 +16,343 @@
...
+        0: {
+          'instances': [
+            {
+              'name': FAKE_INSTANCE.name,
+              'zone': '/' + FAKE_INSTANCE.zone
+            }
+          ]
+        }
+      }
+    }
```
tomchop: As discussed, maybe set these long dicts at the top of the file for better test readability.

libcloudforensics/gcp_test.py
```diff
@@ -16,28 +16,343 @@
...
+      FAKE_INSTANCE.name: {
+        'zone': FAKE_INSTANCE.zone
+      }
+    }
+    found_instance = FAKE_ANALYSIS_PROJECT.get_instance(FAKE_INSTANCE.name, FAKE_INSTANCE.zone)
+    self.assertIsInstance(found_instance, gcp.GoogleComputeInstance)
+    self.assertEqual(found_instance.project, FAKE_INSTANCE.project)
+    self.assertEqual(found_instance.name, FAKE_INSTANCE.name)
+    self.assertEqual(found_instance.zone, FAKE_INSTANCE.zone)
+    self.assertEqual(found_instance._data, FAKE_INSTANCE._data)
```
tomchop: As discussed offline, if you know the value of `FAKE_INSTANCE.name` in advance, it's better to actually have the string literal here rather than a variable. If it's interesting to know which variable it's supposed to be equivalent to, you can add a comment.

libcloudforensics/gcp_test.py
```diff
@@ -16,28 +16,343 @@
...
+    disk_from_snapshot = FAKE_ANALYSIS_PROJECT.create_disk_from_snapshot(FAKE_SNAPSHOT)
+    self.assertIsInstance(disk_from_snapshot, gcp.GoogleComputeDisk)
+    self.assertEqual(disk_from_snapshot.name, self.__get_disk_name_for_snapshot(FAKE_SNAPSHOT))
+
+    # create_disk_from_snapshot(snapshot, disk_name='new-forensics-disk')
+    disk_from_snapshot = FAKE_ANALYSIS_PROJECT.create_disk_from_snapshot(FAKE_SNAPSHOT,
+                                                                         disk_name='new-forensics-disk')
+    self.assertIsInstance(disk_from_snapshot, gcp.GoogleComputeDisk)
+    self.assertEqual(disk_from_snapshot.name, self.__get_disk_name_for_snapshot(FAKE_SNAPSHOT,
+                                                                                 disk_name='new-forensics-disk'))
```
tomchop: Indentation looks off. May I suggest something along the lines of:

```
self.assertEqual(
  disk_from_snapshot.name,
  self.__get_disk_name_for_snapshot(FAKE_SNAPSHOT,disk_name='new-forensics-disk'))
```

Or again, if you know the value in advance, just use that string literal.

libcloudforensics/gcp_test.py
```diff
@@ -16,28 +16,343 @@
...
   def setUp(self):
     super(GoogleCloudProjectTest, self).setUp()
-    self.project = gcp.GoogleCloudProject('test-project')
-    self.project.GceApi = mock.MagicMock()
-    self.project.GceOperation = mock.MagicMock()
 
-  def testFormatLogMessage(self):
-    """Test formatting log message."""
+  def test_format_log_message(self):
+    """Test formatting log message. """
```
tomchop: I'm not sure why a space was added after the `.`, but the style guide recommends ending on-line docstrings by periods. (with no trailing spaces). I think there are some other places where this happens.

libcloudforensics/gcp_test.py
```diff
@@ -16,28 +16,343 @@
...
+    self.assertRaises(RuntimeError, FAKE_ANALYSIS_PROJECT.get_disk, 'non-existent-disk')
+
+  @mock.patch('libcloudforensics.gcp.GoogleCloudProject.gce_operation')
+  @mock.patch('libcloudforensics.gcp.GoogleCloudProject.gce_api')
+  def test_create_disk_from_snapshot(self, mock_gce_api, mock_gce_operation):
+    """Test the creation of a disk from a snapshot. """
+    mock_gce_api.return_value.disks.return_value.insert.return_value.execute.return_value = None
+    mock_gce_operation.return_value = None
+
+    # create_disk_from_snapshot(snapshot)
```
tomchop: If you're gonna add a comment, can you be a bit more explicit? Does this mean that you're just testing this particular function?

libcloudforensics/gcp.py
giovannt0: Addressed

libcloudforensics/gcp_test.py
giovannt0: Addressed

libcloudforensics/gcp.py
giovannt0: Addressed, and I changed all other similar cases for harmonization.

libcloudforensics/gcp.py
giovannt0: Addressed.

libcloudforensics/gcp_test.py
giovannt0: Addressed

libcloudforensics/gcp_test.py
giovannt0: Addressed, where possible

libcloudforensics/gcp_test.py
giovannt0: Addressed. For this one I tried to improve readability. I also ran yapf on the code but I found it to break visibility in some places. Do we want to yapf everything?

libcloudforensics/gcp_test.py
giovannt0: Addressed, other cases as well.

libcloudforensics/gcp_test.py
giovannt0: Addressed. The comment is just here to showcase how the method is called since some parameters are optional. I reflected this more explicitly.

libcloudforensics/gcp.py
```diff
@@ -718,38 +711,37 @@ def add_labels(self, new_labels_dict, blocking_call=False):
     """
 
     get_operation = self.get_operation().execute()
-    labelFingerprint = get_operation['labelFingerprint']
+    label_fingerprint = get_operation['label_fingerprint']
```
tomchop: I think this will break things: https://cloud.google.com/compute/docs/labeling-resources

libcloudforensics/gcp.py
```diff
@@ -718,38 +711,37 @@ def add_labels(self, new_labels_dict, blocking_call=False):
...
+    existing_labels_dict = dict()
     if self.get_labels() is not None:
-      exisitng_labels_dict = self.get_labels()
-    exisitng_labels_dict.update(new_labels_dict)
-    labels_dict = exisitng_labels_dict
-    request_body = {'labels': labels_dict, 'labelFingerprint': labelFingerprint}
+      existing_labels_dict = self.get_labels()
+    existing_labels_dict.update(new_labels_dict)
+    labels_dict = existing_labels_dict
+    request_body = {'labels': labels_dict, 'label_fingerprint': label_fingerprint}
```
tomchop: Same here

libcloudforensics/gcp.py
giovannt0: Fixed

libcloudforensics/gcp.py
giovannt0: Fixed

PR #11, opened 2020-03-12
Completed unittests for gcp.py and added end-to-end test on GCP.

giovannt0:
- Added unittests for: `GoogleComputeBaseResource`, `GoogleComputeInstance`, `GoogleComputeDisk`
- Added unittests for general methods of `gcp.py`
- Added end-to-end test through GCP
- Migrated the tests to a new folder `tests/`


tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+import logging
+import json
+import time
+
+log = logging.getLogger()
+
+
+class EndToEndTest(unittest.TestCase):
+  """End to end test on GCP.
+  Test that gcp.py is able to correctly make a copy of the boot disk and of another given disk
```
tomchop: this line definitely looks like it's over 80 chars.

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+  attached to a given instance. The copies are then attached to another instance to run
+  forensics on the disk copies.
+  Add your project information to a project.info file:
```
tomchop: Feel free to add as much space as needed in docstrings, i.e. it's OK to skip a line if it makes things more readable.

Also, where should the `project.info` file be located? Also, if it's a JSON file it's better to call it `project_info.json` or at least give it the `.json` extension.

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+  """
+
+  def setUp(self):
+    super(EndToEndTest, self).setUp()
+    self.boot_disk_copy = None
+    self.disk_to_forensic_copy = None
+    self.analysis_vm = None
+    self.analysis_vm_name = 'new-vm-for-analysis'
+    try:
+      with open('project.info') as json_file:
```
tomchop: I see. This should probably be an absolute path, passed in e.g. an environment variable? So that the test caller can decide where the file lives.

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+        data = json.load(json_file)
+        self.project_id = data['project_id']
```
tomchop: You can probably "close" the file here after you've opened and loaded it. This would allow you to move the exception management up and group code by functionality

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+        self.instance_to_analyse = data['instance']
+        # Optional: test a disk other than the boot disk
+        if 'disk' in data:
+          self.disk_to_forensic = data['disk']
+        else:
+          self.disk_to_forensic = None
+        self.zone = data['zone']
+    except IOError as exception:
+      raise unittest.SkipTest('Could not set up end to end test: {0:s}'.format(str(exception)))
```
tomchop: Maybe be more explicit here? e.g. `"Could not load the JSON config file from {path}: {0:s}".format(str(error))`

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+  def test_end_to_end(self):
+    """End to end test on GCP.
+    Test that gcp.py is able to correctly make a copy of the boot disk and of another given disk
+    attached to a given instance. The copies are then attached to another instance to run
+    forensics on the disk copies.
+    """
+
+    # Make a copy of the boot disk of the instance to analyse
+    log.info('Boot disk copy started for instance: {0:s}.'
+             .format(self.instance_to_analyse))
```
tomchop: it's better to break the line before the string starts, or before the first parameter of `format`. Same comment for other places where this happens

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+        self.zone = data['zone']
+    except IOError as exception:
+      raise unittest.SkipTest('Could not set up end to end test: {0:s}'.format(str(exception)))
+
+  def test_end_to_end(self):
+    """End to end test on GCP.
+    Test that gcp.py is able to correctly make a copy of the boot disk and of another given disk
+    attached to a given instance. The copies are then attached to another instance to run
+    forensics on the disk copies.
+    """
```
tomchop: It would be really cool if this docstring described a clear, numbered sequence of things happening.

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+      cpu_cores=4,
+      attach_disk=self.boot_disk_copy
+    )
+    log.info('Instance {0:s} started.'
+             .format(self.analysis_vm_name))
+    self.assertIsInstance(self.analysis_vm, gcp.GoogleComputeInstance)
+    self.assertTrue(created)
+    self.assertEqual(self.analysis_vm.name, 'new-vm-for-analysis')
+
+    if self.disk_to_forensic is not None:
```
tomchop: Tip: use a guard statement and save one indentation level in all lines below!

```python
if self.disk_to_forensic is None:
  return
# Make a copy of another disk of the instance to analyse
# [snip]
```

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+    log.warning(disks)
+    if self.disk_to_forensic is not None:
+      self.assertEqual(disks,
+                       ['new-vm-for-analysis',
+                        self.boot_disk_copy.name,
+                        self.disk_to_forensic_copy.name])
+    else:
+      self.assertEqual(disks,
+                       ['new-vm-for-analysis',
+                        self.boot_disk_copy.name])
```
tomchop: How is this ever going to be reached if you check for the opposite condition in L102?

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+             .format(self.analysis_vm.name))
+    operation = project.gce_api().instances().delete(
+      project=project.project_id,
+      zone=self.zone,
+      instance=self.analysis_vm.name
+    ).execute()
+    try:
+      project.gce_operation(operation, block=True)
+    except HttpError:
+      pass
```
tomchop: Please add a comment explaining why we don't care about the exception here.

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+    log.info('Instance {0:s} successfully deleted.'
+             .format(self.analysis_vm.name))
+
+    # We wait 5min before trying to delete the disk so that GCE propagates the previous operation,
+    # as the instance first needs to be stopped, and then deleted.
+    time.sleep(300)
```
tomchop: Ah, I remember discussing this. It's OK for now, but what would be /better/ would be to busy-loop around the existence or not of the VM that we're waiting to be deleted

tests/e2e.py
```diff
@@ -0,0 +1,204 @@
...
+    ).execute()
+    try:
+      project.gce_operation(operation, block=True)
+    except HttpError:
+      pass
+    log.info('Disk {0:s} successfully deleted.'
+             .format(self.boot_disk_copy.name))
+
+    if self.disk_to_forensic is not None:
+      log.info('Deleting disk: {0:s}.'
```
tomchop: Same comment about guard statements

PR #12, opened 2020-03-18
Refactor e2e test and add Codecov to github actions.

giovannt0:

### DIFF
--- a/.github/workflows/install_test_setup_py.yml
+++ b/.github/workflows/install_test_setup_py.yml
@@ -23,3 +23,13 @@ jobs:
       run: python setup.py install
     - name: Test through setup.py
       run: python setup.py test
+    - name: Install coverage
+      run: pip install coverage
+    - name: Create coverage report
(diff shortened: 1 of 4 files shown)


giovannt0:
Takes care of #10 as well.

.github/workflows/install_test_setup_py.yml
```diff
@@ -23,3 +23,13 @@ jobs:
       run: python setup.py install
     - name: Test through setup.py
       run: python setup.py test
+    - name: Install coverage
+      run: pip install coverage
+    - name: Create coverage report
+      run: coverage run --omit '/usr/*' ./setup.py test && coverage xml
```
tomchop: This is gonna run tests twice again, if I understand correctly. Maybe we can skip the test run in L 24 / 25?

Also, this will run coverage reports for all versions of python as defined in the matrix (so about 6 times). Maybe having a separate action to run coverage would be simpler in this case.

giovannt0: Correct; coverage migrated to its own GitHub action.

PR #17, opened 2020-03-24
Fix disk name generation.

### LINKED ISSUE #16: Fix disk name creation
There's a bug in `GoogleCloudProject.crreate_disk_from_snapshot()` with the disk's name creation. If the project name and the instance name are long strings, then `truncate_at` becomes negative and the resulting `disk_name` that is created may end up being longer than 63 chars. Additionally, `len(project_id)` should not be taken out of `truncate_at` since the project id itself is not used in the disk name. 

The disk name creation code should be factored out of this method, and unit tested separately.

giovannt0:
Fixes #16 

### DIFF
--- a/libcloudforensics/gcp.py
+++ b/libcloudforensics/gcp.py
@@ -25,6 +25,7 @@ import datetime
 import json
 import logging
 import os
+import re
 import socket
 import ssl
 import subprocess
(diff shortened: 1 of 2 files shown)


giovannt0:
Good catch with RFC1035. Turns out there were multiple places in the code where disk/snapshot names built from un-verified user inputs could result in a violation of the RegEx. Therefore, the latest commit also adds a check where necessary when creating disks / snapshots from user-passed parameter.

Issue #13, opened 2020-03-20
Refactor e2e tests to be consistent with dftimewolf


tomchop:
See this PR for more details https://github.com/log2timeline/dftimewolf/pull/257/files

- [x] Folder structure (unittests / e2e tests)
- [x] Function to check validity of JSON file

 #19

PR #19, opened 2020-03-24
Make e2e test consistent with dfTimewolf.

giovannt0:


giovannt0:
Second task of #13 can be taken care of once #17 is merged.

PR #25, opened 2020-03-30
Exported bash script to its own file

### LINKED ISSUE #24: Extract bash script to file
Factor out the bash script inside `gcp.py` into its own file.

giovannt0:
Closes #24 

### DIFF
--- a/libcloudforensics/gcp.py
+++ b/libcloudforensics/gcp.py
@@ -41,6 +41,7 @@ log = logging.getLogger()
 
 RETRY_MAX = 10
 REGEX_DISK_NAME = re.compile('^(?=.{1,63}$)[a-z]([-a-z0-9]*[a-z0-9])?$')
+STARTUP_SCRIPT = 'scripts/startup.sh'
 
 
 def CreateService(service_name, api_version):
@@ -327,7 +328,8 @@ class GoogleCloudProject:
 
   def GetOrCreateAnalysisVm(
       self, vm_name, boot_disk_size, cpu_cores=4,
-      image_project='ubuntu-os-cloud', image_family='ubuntu-1804-lts'):
+      image_project='ubuntu-os-cloud', image_family='ubuntu-1804-lts',
+      packages=None):
     """Get or create a new virtual machine for analysis purposes.
 
     Args:
@@ -337,6 +339,7 @@ class GoogleCloudProject:
       image_project (str): Name of the project where the analysis VM image is
       hosted.
       image_family (str): Name of the image to use to create the analysis VM.
+      packages (list(str)): List of packages to install in the VM
 
     Returns:
       tuple(GoogleComputeInstance, bool): A tuple with a virtual machine object
@@ -362,6 +365,11 @@ class GoogleCloudProject:
         project=image_project, family=image_family).execute()
     ubuntu_image = self.GceOperation(get_image_operation, block=False)
     source_disk_image = ubuntu_image['selfLink']
+
+    startup_script = self._ReadStartupScript()
+
+    if packages:
+      startup_script.replace('${packages[@]}', ' '.join(packages))
 
     config = {
         'name': vm_name,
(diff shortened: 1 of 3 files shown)


tests/unittest/gcp_test.py
```diff
@@ -418,6 +419,35 @@ def testListDisksByLabels(self, mock_gce_api, mock_gce_operation):
...
+
+    # Bogus environment variable, should raise an exception
+    os.environ['STARTUP_SCRIPT'] = '/bogus/path'
+    self.assertRaises(OSError, FAKE_SOURCE_PROJECT._ReadStartupScript)
+
+    # Environment variable set to custom script
+    file = """#!/bin/bash
+    echo 'This is a custom script'
+    """
+    file_path = os.getcwd() + '/tmp.sh'
```
tomchop: Better use a NamedTemporaryFile for this: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

scripts/startup.sh
```diff
@@ -0,0 +1,43 @@
...
+)
+
+err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $*" >&2
+}
+
+install_packages() {
+  add-apt-repository -y -u ppa:gift/${gift_ppa_track}
+  for package in ${packages[@]}; do
+    apt -y install ${package}
```
tomchop: It would maybe be faster if we installed them all in one go?

libcloudforensics/gcp.py
```diff
@@ -534,6 +504,35 @@ def __ListByLabel(self, labels_filter, service_object, filter_union):
...
+
+    Raises:
+      OSError: If the script cannot be opened, read or closed.
+    """
+    try:
+      startup_script = os.environ.get('STARTUP_SCRIPT')
+      if not startup_script:
+        # Use the provided script
+        startup_script = os.path.dirname(os.path.realpath(__file__)) + \
+                         '/../scripts/startup.sh'
```
tomchop: Use `os.path.dirname` to go up a level rather than `..`: https://docs.python.org/3/library/os.path.html#os.path.dirname (you can use it as many times as you want to go up as many levels as you want)

scripts/startup.sh
giovannt0: Done. To address your overall comment I've also added a new argument to the `GetOrCreateAnalysisVm` to specify a particular list of packages to install.

tests/unittest/gcp_test.py
```diff
@@ -435,15 +436,14 @@ def testReadStartupScript(self):
...
-      script = FAKE_SOURCE_PROJECT._ReadStartupScript()
-      self.assertEqual(script, file)
-      os.remove(file_path)
+      with tempfile.NamedTemporaryFile() as temp:
+        temp.write(str.encode(file))
+        temp.flush()
+        os.environ['STARTUP_SCRIPT'] = temp.name
+        script = FAKE_SOURCE_PROJECT._ReadStartupScript()
+        self.assertEqual(script, file)
+        temp.close()
```
tomchop, line 446: No need to call `close()` if you're within a `with` statement, the cleanup is done automatically (even if an exception is raised)

libcloudforensics/gcp.py
```diff
@@ -523,8 +530,8 @@ def _ReadStartupScript():
       startup_script = os.environ.get('STARTUP_SCRIPT')
       if not startup_script:
         # Use the provided script
-        startup_script = os.path.dirname(os.path.realpath(__file__)) + \
-                         '/../scripts/startup.sh'
+        startup_script = os.path.dirname(os.path.dirname(os.path.realpath(
+            __file__))) + '/scripts/startup.sh'
```
tomchop, line 534: Ah yes, I guess I also should've mentioned to use [os.path.join()](https://docs.python.org/3.7/library/os.path.html#os.path.join) here as well.

giovannt0:
Thanks, addressed!

PR #28, opened 2020-04-07
Add snapshoting capabilities on AWS.

giovannt0:
Similar to capabilities in `gcp.py`, this PR brings support for snapshoting volumes on AWS. `CreateVolumeCopy(...)` bootstraps a new volume created from a snapshot of another volume.

### DIFF
--- a/libcloudforensics/aws.py
+++ b/libcloudforensics/aws.py
@@ -77,7 +77,7 @@ class AWSAccount:
         service_name=service, region_name=self.default_region
     )
 
-  def ListInstances(self, region=None, filters=None):
+  def ListInstances(self, region=None, filters=None, show_terminated=False):
     """List instances of an AWS account.
 
     Example usage:
@@ -87,6 +87,8 @@ class AWSAccount:
     Args:
       region (str): Optional. The region from which to list instances.
       filters (list(dict)): Optional. Filters for the query.
+      show_terminated (bool): Optional. Include terminated instances in the
+      list.
 
     Returns:
       dict: Dictionary with name and metadata for each instance.
@@ -114,10 +116,9 @@ class AWSAccount:
 
       for reservation in response['Reservations']:
         for instance in reservation['Instances']:
-          # Terminated instances are filtered out. If
-          # reservation['Instances'] contains any entry, then the
+          # If reservation['Instances'] contains any entry, then the
           # instance's state is expected to be present in the API's response.
-          if instance['State']['Name'] == 'terminated':
+          if instance['State']['Name'] == 'terminated' and not show_terminated:
             continue
 
           zone = instance['Placement']['AvailabilityZone']
--- a/libcloudforensics/aws.py
+++ b/libcloudforensics/aws.py
@@ -38,8 +38,7 @@ class AWSAccount:
   def __init__(self, default_availability_zone):
     self.default_availability_zone = default_availability_zone
     # The region is given by the zone minus the last letter
-    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions
-    # -availability-zones.html#using-regions-availability-zones-describe
+    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#using-regions-availability-zones-describe # pylint: disable=line-too-long
     self.default_region = self.default_availability_zone[:-1]
 
   def ClientApi(self, service, region=None):
@@ -81,6 +80,10 @@ class AWSAccount:
   def ListInstances(self, region=None, filters=None):
     """List instances of an AWS account.
 
+    Example usage:
+      ListInstances(region='us-east-1', filters=[
+          {'Name':'instance-id', 'Values':['some-instance-id']}])
+
     Args:
       region (str): Optional. The region from which to list instances.
       filters (list(dict)): Optional. Filters for the query.
@@ -90,10 +93,6 @@ class AWSAccount:
 
     Raises:
       RuntimeError: If instances can't be listed.
-
-    Example usage:
-      ListInstances(region='us-east-1', filters=[
-          dict(Name='instance-id', Values=['some-instance-id'])])
     """
     if not filters:
       filters = []
@@ -110,7 +109,7 @@ class AWSAccount:
         else:
           response = client.describe_instances(Filters=filters)
       except client.exceptions.ClientError as exception:
-        raise RuntimeError('Error: could not retrieve instances: {0:s}'.format(
+        raise RuntimeError('Could not retrieve instances: {0:s}'.format(
             str(exception)))
 
       for reservation in response['Reservations']:
@@ -143,6 +142,11 @@ class AWSAccount:
   def ListVolumes(self, region=None, filters=None):
     """List volumes of an AWS account.
 
+    Example usage:
+      # List volumes attached to the instance 'some-instance-id'
+      ListVolumes(filters=[
+          {'Name':'attachment.instance-id', 'Values':['some-instance-id']}])
+
     Args:
       region (str): Optional. The region from which to list the volumes.
       filters (list(dict)): Optional. Filter for the query.
@@ -152,11 +156,6 @@ class AWSAccount:
 
     Raises:
       RuntimeError: If volumes can't be listed.
-
-    Example usage:
-      # List volumes attached to the instance 'some-instance-id'
-      ListVolumes(filters=[
-          dict(Name='attachment.instance-id', Values=['some-instance-id'])])
     """
     if not filters:
       filters = []
@@ -173,7 +172,7 @@ class AWSAccount:
         else:
           response = client.describe_volumes(Filters=filters)
       except client.exceptions.ClientError as exception:
-        raise RuntimeError('Error: could not retrieve volumes: {0:s}'.format(
+        raise RuntimeError('Could not retrieve volumes: {0:s}'.format(
             str(exception)))
 
       for volume in response['Volumes']:
@@ -223,7 +222,7 @@ class AWSAccount:
           are set.
     """
     if not instance_name and not instance_id or instance_name and instance_id:
-      raise ValueError('Error: you must specify exactly one of [instance_name, '
+      raise ValueError('You must specify exactly one of [instance_name, '
                        'instance_id]. Got instance_name: {0:s}, instance_id: '
                        '{1:s}'.format(instance_name, instance_id))
     if instance_name:
@@ -254,7 +253,7 @@ class AWSAccount:
           are set.
     """
     if not volume_name and not volume_id or volume_name and volume_id:
-      raise ValueError('Error: you must specify exactly one of [volume_name, '
+      raise ValueError('You must specify exactly one of [volume_name, '
                        'volume_id]. Got volume_name: {0:s}, volume_id: '
                        '{1:s}'.format(volume_name, volume_id))
     if volume_name:
@@ -284,7 +283,7 @@ class AWSAccount:
           snapshot, volume_name_prefix=volume_name_prefix)
 
     if not REGEX_TAG_VALUE.match(volume_name):
-      raise ValueError('Error: volume name {0:s} does not comply with '
+      raise ValueError('Volume name {0:s} does not comply with '
                        '{1:s}'.format(volume_name, REGEX_TAG_VALUE.pattern))
 
     client = self.ClientApi(EC2_SERVICE)
@@ -300,7 +299,7 @@ class AWSAccount:
       client.get_waiter('volume_available').wait(VolumeIds=[volume_id])
     except (client.exceptions.ClientError,
             botocore.exceptions.WaiterError) as exception:
-      raise RuntimeError('Error: could not create volume {0:s} from snapshot '
+      raise RuntimeError('Could not create volume {0:s} from snapshot '
                          '{1:s}: {2:s}'.format(volume_name, snapshot.name,
                                                str(exception)))
 
@@ -323,8 +322,7 @@ class AWSAccount:
 
     # Max length of tag values in AWS is 255 characters
     # UserId is expected to be set if the call to the EC2 API is successful.
-    # See https://boto3.amazonaws.com/v1/documentation/api/1.9.42/reference/
-    # services/sts.html#STS.Client.get_caller_identity for more details.
+    # https://boto3.amazonaws.com/v1/documentation/api/1.9.42/reference/services/sts.html#STS.Client.get_caller_identity for more details. # pylint: disable=line-too-long
     user_id = self.ClientApi(ACCOUNT_SERVICE).get_caller_identity()['UserId']
     volume_id = user_id + snapshot.volume.volume_id
     volume_id_crc32 = '{0:08x}'.format(
@@ -581,7 +579,7 @@ class AWSVolume(AWSElasticBlockStore):
     truncate_at = 255 - len(timestamp) - 1
     snapshot_name = '{0}-{1}'.format(snapshot_name[:truncate_at], timestamp)
     if not REGEX_TAG_VALUE.match(snapshot_name):
-      raise ValueError('Error: snapshot name {0:s} does not comply with '
+      raise ValueError('Snapshot name {0:s} does not comply with '
                        '{1:s}'.format(snapshot_name, REGEX_TAG_VALUE.pattern))
 
     client = self.aws_account.ClientApi(EC2_SERVICE)
@@ -595,7 +593,7 @@ class AWSVolume(AWSElasticBlockStore):
       client.get_waiter('snapshot_completed').wait(SnapshotIds=[snapshot_id])
     except (client.exceptions.ClientError,
             botocore.exceptions.WaiterError) as exception:
-      raise RuntimeError('Error: could not create snapshot for volume {0:s}: '
+      raise RuntimeError('Could not create snapshot for volume {0:s}: '
                          '{1:s}'.format(self.volume_id, str(exception)))
 
     return AWSSnapshot(snapshot_id, self, name=snapshot_name)
@@ -628,7 +626,7 @@ class AWSSnapshot(AWSElasticBlockStore):
     try:
       client.delete_snapshot(SnapshotId=self.snapshot_id)
     except client.exceptions.ClientError as exception:
-      raise RuntimeError('Error: could not delete snapshot {0:s}: {1:s}'.format(
+      raise RuntimeError('Could not delete snapshot {0:s}: {1:s}'.format(
           self.snapshot_id, str(exception)))
 
 
@@ -667,7 +665,7 @@ def CreateVolumeCopy(instance_id, zone, volume_id=None):
         volume_to_copy.volume_id, new_volume.volume_id))
 
   except RuntimeError as exception:
-    error_msg = 'Error copying volume {0:s}: {1!s}'.format(
+    error_msg = 'Copying volume {0:s}: {1!s}'.format(
         volume_id, exception)
     raise RuntimeError(error_msg)
 
--- a/libcloudforensics/aws.py
+++ b/libcloudforensics/aws.py
@@ -1,0 +1,726 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Library for incident response operations on AWS EC2."""
+import binascii
+import datetime
+import logging
+import os
+import re
+
+import boto3
+
+log = logging.getLogger()
+
+EC2_SERVICE = 'ec2'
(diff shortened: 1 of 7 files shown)


libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+  # /configuration.html
+  return boto3.session.Session()
+
+
+class AwsAccount:
+  """Class representing an AWS account.
+
+  Attributes:
+    default_availability_zone (str): Default zone within the region to create
+    new resources in.
```
tomchop, line 51: Add missing indentation spaces

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+def CreateSession():
+  """Create an AWS session API service.
+
+  Returns:
+    boto3.Session: An AWS session object.
+  """
+  # Create a default session with credentials stored in ~/.aws/credentials
+  # https://boto3.amazonaws.com/v1/documentation/api/latest/guide
+  # /configuration.html
+  return boto3.session.Session()
```
tomchop, line 43: This wrapper is VERY thin, is it necessary? It seems like we could just call `boto3.session.Session()` directly in the rest of the code?

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+        response = self.ClientApi(
+            EC2_SERVICE, region=region).describe_instances(
+                Filters=filters,
+                NextToken=next_token)
+      else:
+        response = self.ClientApi(
+            EC2_SERVICE, region=region).describe_instances(
+                Filters=filters)
+
+      for reservation in response["Reservations"]:
```
tomchop, line 129: Please use single quotes for strings here and below

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+        for instance in reservation["Instances"]:
+          instance_name = None
+          if instance.get('Tags'):
+            for tag in instance['Tags']:
```
tomchop, line 133: You can replace this by 

```python
for tag in instance.get('Tags', []):
```

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+              instance_name = tag.get('Value') if tag.get('Key') == 'Name' \
+                else None
+              if instance_name is not None:
+                break
```
tomchop, line 137: Do you want to stop processing all instances here? Or skip to the next instance (if the latter, use `continue`)

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+            EC2_SERVICE, region=region).describe_instances(
+                Filters=filters)
+
+      for reservation in response["Reservations"]:
+        for instance in reservation["Instances"]:
+          instance_name = None
+          if instance.get('Tags'):
+            for tag in instance['Tags']:
+              instance_name = tag.get('Value') if tag.get('Key') == 'Name' \
+                else None
```
tomchop, line 135: `instance_name` is already set to None in L131. I would suggest the following change:

```python
if tag.get('Key') == 'Name':
  instance_name = tag.get('Value')
if instance_name is not None:
  break # or continue, see comment below.
```

Also, we may want to use `if instance_name:` without the `is not None:`. Using the latter means that you wanna treat empty* values (like `''`, `{}`, `[]`) in a different way than if the `Value` key is not set.

\* or values that evaluate to `False` when converted to a `bool`

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+              if instance_name is not None:
+                break
+          # Terminated instances are filtered out
+          if instance['State']['Name'] != 'terminated':
```
tomchop, line 139:
```python
if instance['State']['Name'] == 'terminated':
  continue
```

Remember, guard statements are your friends :) 

Also, I see you're very careful in the lines before this regarding the expected keys to be set, but not here. Is there a reason why?
If the code is never functional when keys are not present, it's OK to raise an exception that can be caught by the caller.

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+            instance_info = dict(region=self.default_region,
+                                 zone=instance['Placement']['AvailabilityZone'])
+            if instance_name:
+              instance_info.update(name=instance_name)
```
tomchop, line 143: You can also do `instance_info[name] = instance_name`

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+            instances[instance['InstanceId']] = instance_info
+
+      next_token = response.get('NextToken')
+      if not next_token:
+        have_all_tokens = True
```
tomchop, line 148: I'm sure you can use some kind of `break` statement here in your loop

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+    Example usage:
+      # List volumes attached to the instance 'a-particular-instance-id'
+      ListVolumes(filters=[
+        dict(Name='attachment.instance-id', Values=['a-particular-instance-id'])
+      ])
+    """
+    if not filters:
+      filters = []
+
+    volumes = dict()
```
tomchop, line 171: Thought for a future PR: it would be nice to make the `dict` creation call consistent. I see `dict()` and `{}` being used in the codebase

libcloudforensics/aws.py
```diff
@@ -0,0 +1,726 @@
...
+import os
+import re
+
+import boto3
+
+log = logging.getLogger()
+
+EC2_SERVICE = 'ec2'
+ACCOUNT_SERVICE = 'sts'
+UBUNTU_1804_AMI = 'ami-0013b3aa57f8a4331'
```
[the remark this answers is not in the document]
giovannt0: Ah, actually this is for next PR.I'll add the description there. Removed in this one.

libcloudforensics/aws.py
giovannt0: This is not consistent in `gcp.py` either, I'll fix it there. I would rather go with the `dict()` call, wdyt?

libcloudforensics/aws.py
giovannt0: If there's an instance in `reservation['Instances']`, then the EC2 API always include the state of that instance, among other informations (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_instances). It's therefore expected that these keys are set if the API call was successful. However I realised that there was no try/except statement around this very API call. This has been added.

libcloudforensics/aws.py
giovannt0: We stop processing the tags as they're unique and we're only interested in the `Name` key

libcloudforensics/aws.py
tomchop: Please add a comment saying that if the above exists, then these keys can be expected to be set.

libcloudforensics/aws.py
tomchop: I'd go for the literal (`{}`, or `[]` if you're using lists), it's clearer than a function call, and according to the Internet, it's also a bit faster (but that's really bold coming from Python :P)

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+    for key in volumes:
+      if volumes[key].get('name', None) == volume_name:
+        volume_id = key
+        count += 1
+
+    if count == 0:
+      error_msg = 'Volume {0:s} was not found in AWS account'.format(
+          volume_name)
+      raise RuntimeError(error_msg)
+    if count > 1:
```
tomchop, line 449: Same comment as in `__GetInstanceByName`

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+    region (str): The region the instance is in.
+    availability_zone (str): The zone within the region in which the instance
+        is.
+    name (str): the name tag (if any) of the instance.
+  """
+  def __init__(self, aws_account, instance_id, region,
+               availability_zone, name=None):
+    """Initialize the AWS EC2 instance.
+
+    Attributes:
```
tomchop, line 478: I think you want `Args:` here

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+    super(AWSSnapshot, self).__init__(volume.aws_account, volume.region,
+                                      volume.availability_zone, name)
+    self.snapshot_id = snapshot_id
+    self.volume = volume
+
+  def Delete(self):
+    """Delete a snapshot."""
+    client = self.aws_account.ClientApi(EC2_SERVICE)
+    try:
+      client.delete_snapshot(
```
tomchop, line 639: Doesn't this fit in one line?

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+          SnapshotId=self.snapshot_id)
+    except client.exceptions.ClientError as exception:
+      raise RuntimeError('Error: could not delete snapshot {0:s}: {1:s}'.format(
+          self.snapshot_id, str(exception)))
+
+
+def CreateVolumeCopy(instance_name, zone, volume_name=None):
+  """Create a copy of an AWS EBS Volume.
+
+  Attributes:
```
tomchop, line 649: `Args:` for functions, `Attributes:` for classes

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+  """
+  return dict(
+      ResourceType=resource,
+      Tags=[
+          dict(
+              Key='Name',
+              Value=name
+          )
+      ]
+  )
```
tomchop, line 705: Using dict litterals, this would look like

```python
return {
  'ResourceType': resource,
  'Tags': [{
    'Key': 'Name',
    'Value': name,
  }],
}
```

which IMO is much nicer to read.

libcloudforensics/aws.py
```diff
@@ -103,49 +87,50 @@ def ListInstances(self, region=None, filters=None):
...
-            if instance_name:
-              instance_info.update(name=instance_name)
-            instances[instance['InstanceId']] = instance_info
+          if instance['State']['Name'] == 'terminated':
+            continue
+
+          instance_info = dict(region=self.default_region,
+                               zone=instance['Placement']['AvailabilityZone'])
+
+          for tag in instance.get('Tags', []):
```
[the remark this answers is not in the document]
giovannt0: Correct. Since EC2 instances are not referenced to by their names like in GCP but only their IDs, adding a name to any instance is completely optional for users. If none is assigned to an instance, then `instance_info` will only contain region and zone information.

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+
+        volumes[volume['VolumeId']] = volume_info
+
+      next_token = response.get('NextToken')
+      if not next_token:
+        break
+
+    return volumes
+
+  def GetInstance(self, instance_name_or_id, region=None):
```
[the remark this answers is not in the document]
giovannt0: Yeah, this is much cleaner. I made the methods public and left `GetInstanceByNameOrId` as a convenience method for users. Same goes for retrieving volumes.

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+      volume = client.create_volume(
+          AvailabilityZone=snapshot.availability_zone,
+          SnapshotId=snapshot.snapshot_id,
+          TagSpecifications=[GetTagForResourceType('volume', volume_name)]
+      )
+      volume_id = volume['VolumeId']
+      zone = volume['AvailabilityZone']
+      # Wait for volume creation completion
+      client.get_waiter('volume_available').wait(VolumeIds=[volume_id])
+    except client.exceptions.ClientError as exception:
```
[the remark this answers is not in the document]
giovannt0: `get_waiter()` may throw a `ValueError` if the waiter is not found. The method's parameter is however not under user's control so if it breaks it's likely due to a change within boto3. `wait()` on the other hand can throw a `botocore.exceptions.WaiterError`. I added this error in the same catch block as of now, and updated other similar parts of the code . Do you think it's worth making it two separate try/except blocks?

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+    Returns:
+      str: A name for the volume.
+
+    Raises:
+      ValueError: if the volume name does not comply with the RegEx.
+    """
+
+    # Max length of tag values in AWS is 255 characters
+    user_id = self.ClientApi(ACCOUNT_SERVICE).get_caller_identity()\
+      .get('UserId', '')
```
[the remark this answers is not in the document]
giovannt0: Removed `\`.  According to https://boto3.amazonaws.com/v1/documentation/api/1.9.42/reference/services/sts.html#STS.Client.get_caller_identity, `get_caller_identity()` returns details about the IAM identity whose credentials are used to call the API. If the call is successful, then `UserId` is set to be the unique identifier of the calling entity. My understanding is that if the call succeeds, then this value is guaranteed to be set and unique. I removed the `get()` call and replaced it by ['UserId'] accordingly, and made it clear in the code by adding a comment.

libcloudforensics/aws.py
```diff
@@ -0,0 +1,705 @@
...
+    for key in instances:
+      if instances[key].get('name') == instance_name:
+        instance_id = key
+        count += 1
+
+    if count == 0:
+      error_msg = 'Instance {0:s} was not found in AWS account'.format(
+          instance_name)
+      raise RuntimeError(error_msg)
+    if count > 1:
```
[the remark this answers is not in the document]
giovannt0: Ah, yeah this is great. I also removed the raised errors in favour of returning an empty list of no matching instances/volumes with matching name tag could be found; and adapted `GetInstancesByNameOrId` and `GetVolumesByNameOrId` accordingly.

giovannt0:
Very helpful, thank you; addressed.

libcloudforensics/aws.py
```diff
@@ -0,0 +1,695 @@
...
+  """Class representing an AWS account.
+
+  Attributes:
+    default_availability_zone (str): Default zone within the region to create
+        new resources in.
+  """
+  def __init__(self, default_availability_zone):
+    self.default_availability_zone = default_availability_zone
+    # The region is given by the zone minus the last letter
+    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions
```
rgayon, line 41: No need to split URLS

libcloudforensics/aws.py
```diff
@@ -0,0 +1,695 @@
...
+      region (str): Optional. The region from which to list instances.
+      filters (list(dict)): Optional. Filters for the query.
+
+    Returns:
+      dict: Dictionary with name and metadata for each instance.
+
+    Raises:
+      RuntimeError: If instances can't be listed.
+
+    Example usage:
```
rgayon, line 94: Love these examples!
Maybe move them after the first description line?

libcloudforensics/aws.py
```diff
@@ -0,0 +1,695 @@
...
+              'region': zone[:-1],
+              'zone': zone
+          }
+
+          for tag in instance.get('Tags', []):
+            if tag.get('Key') == 'Name':
+              instance_info['name'] = tag.get('Value')
+              break
+
+          instances[instance['InstanceId']] = instance_info
```
rgayon, line 135: Why not use AWSInstance as values in the dict "instances"?

libcloudforensics/aws.py
```diff
@@ -0,0 +1,695 @@
...
+              Filters=filters, NextToken=next_token)
+        else:
+          response = client.describe_instances(Filters=filters)
+      except client.exceptions.ClientError as exception:
+        raise RuntimeError('Error: could not retrieve instances: {0:s}'.format(
+            str(exception)))
+
+      for reservation in response['Reservations']:
+        for instance in reservation['Instances']:
+          # Terminated instances are filtered out. If
```
rgayon, line 118: Are we sure we want this? Could this be a flag to ListInstance? like terminated=False

libcloudforensics/aws.py
```diff
@@ -0,0 +1,695 @@
...
+
+    Returns:
+      list(AWSInstance): A list of Amazon EC2 Instance objects.
+
+    Raises:
+      ValueError: If both instance_name and instance_id are None or if both
+          are set.
+    """
+    if not instance_name and not instance_id or instance_name and instance_id:
+      raise ValueError('Error: you must specify exactly one of [instance_name, '
```
rgayon, line 226: Probably can remove 'Error: ' in the string message of a ValueError exception =)

rgayon: Same thing at other lines

libcloudforensics/aws.py
giovannt0: The original thought was to be as consistent as possible with the methods' responses that `gcp.py` defined. Should we adapt both, or leave it as is?

libcloudforensics/aws.py
giovannt0: In what use case would users want to list instances that are terminated? Terminated instances cannot be started again and usually disappear from a user's AWS account within a few hours. If you terminate an instance and use boto3's API, it still shows up in the API's response for that period of time, which was why it was being filtered out initially.

giovannt0:
Thanks @rgayon, addressed!

libcloudforensics/aws.py
rgayon: Oh I see.
Maybe add a TODO or a new issue to consider having some kind of generic Instance and Disk interface/object could be nice, since all of these basically have names, zones, regions, etc

rgayon: https://github.com/google/cloud-forensics-utils/issues/32

libcloudforensics/aws.py
rgayon: okay, I thought "terminated" meant "stopped".
I'd still like to be able to collect these,  we've had a couple cases when after asking for someone to stop an instance, they actually deleted it in GCP

giovannt0: Added :)

libcloudforensics/aws.py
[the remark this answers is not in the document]
tomchop: Ack, thanks!

libcloudforensics/aws.py
```diff
@@ -0,0 +1,694 @@
...
+        volumes[volume['VolumeId']] = volume_info
+
+      next_token = response.get('NextToken')
+      if not next_token:
+        break
+
+    return volumes
+
+  def GetInstancesByNameOrId(self, instance_name=None,
+                             instance_id=None, region=None):
```
tomchop, line 202: Please use one parameter per line (or all on the same line when possible) for better readability

libcloudforensics/aws.py
```diff
@@ -0,0 +1,694 @@
...
+    Exactly one of [instance_name, instance_id] must be specified. If looking up
+    an instance by its ID, the method returns a list with exactly one
+    element. If looking up instances by their name tag (which are not unique
+    across instances), then the method will return a list of all instances
+    with that name tag, or an empty list if no instances with matching name
+    tag could be found.
+
+    Args:
+      instance_name (str): Optional. The instance name tag of the instance to
+          get.
```
tomchop, line 214: Please double-check other docstrings in the codebase are not all are indented when more than one line

libcloudforensics/aws.py
```diff
@@ -0,0 +1,694 @@
...
+      region (str): Optional. The region to look the instance in.
+
+    Returns:
+      list(AWSInstance): A list of Amazon EC2 Instance objects.
+
+    Raises:
+      ValueError: If both instance_name and instance_id are None or if both
+          are set.
+    """
+    if not instance_name and not instance_id or instance_name and instance_id:
```
tomchop, line 225: I would add some parenthesis here just to make the boolean statement a bit clearer

libcloudforensics/aws.py
```diff
@@ -0,0 +1,694 @@
...
+          snapshot.name[:truncate_at], volume_id_crc32)
+
+    return volume_name
+
+  def GetInstanceById(self, instance_id, region=None):
+    """Get an instance from an AWS account by its ID.
+
+    Args:
+      instance_id (str): The instance id.
+      region (str): Optional. The region to look the instance in.
```
tomchop, line 351: What happens when the caller doesn't specify a region? Will it look in all regions? Can more than one instance be returned in that case? (or do instances have a unique ID per region?). Same question for other functions

libcloudforensics/aws.py
```diff
@@ -0,0 +1,694 @@
...
+
+    error_msg = 'Boot volume not found for instance: {0:s}'.format(
+        self.instance_id)
+    raise RuntimeError(error_msg)
+
+  def ListVolumes(self):
+    """List all volumes for the instance.
+
+    Returns:
+      dict: Dict of volume ids.
```
tomchop, line 504: This should be consistent with the return doc of `AWSAccount.ListVolumes`

libcloudforensics/aws.py
giovannt0: https://github.com/google/cloud-forensics-utils/issues/34

libcloudforensics/aws.py
giovannt0: If not specified it will use `self.default_region` which is derived from `self.default_availability_zone`, that is given by the user when instantiating `AWSAccount`. It will therefore always look in one region only, so it'll return only one instance if the ID exists in that region. This is a limitation of the `boto3` API, as instances of `boto3.session.Session().client()` must specify exactly one region.

giovannt0:
@tomchop thanks!

libcloudforensics/aws.py
tomchop: Ok I see. Could you please add this to the docstring?

PR #38, opened 2020-04-25
Add unittests for AWS snapshoting capabilities.

giovannt0:

### DIFF
--- a/tests/unittest/aws_test.py
+++ b/tests/unittest/aws_test.py
@@ -152,14 +152,15 @@ class AWSAccountTest(unittest.TestCase):
   @mock.patch('libcloudforensics.aws.AWSAccount.ClientApi')
   def testListInstances(self, mock_ec2_api):
     """Test that instances of an account are correctly listed."""
-    mock_ec2_api.return_value.describe_instances.return_value = MOCK_DESCRIBE_INSTANCES  # pylint: disable=line-too-long
+    describe_instances = mock_ec2_api.return_value.describe_instances
+    describe_instances.return_value = MOCK_DESCRIBE_INSTANCES
     instances = FAKE_AWS_ACCOUNT.ListInstances()
     self.assertEqual(1, len(instances))
     self.assertIn('fake-instance-id', instances)
     self.assertEqual('fake-zone-2', instances['fake-instance-id']['region'])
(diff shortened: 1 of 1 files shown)


tests/unittest/aws_test.py
```diff
@@ -0,0 +1,379 @@
...
+}
+
+
+class AWSAccountTest(unittest.TestCase):
+  """Test AWSAccount class."""
+
+  @mock.patch('libcloudforensics.aws.AWSAccount.ClientApi')
+  def testListInstances(self, mock_ec2_api):
+    """Test that instances of an account are correctly listed."""
+    mock_ec2_api.return_value.describe_instances.return_value = MOCK_DESCRIBE_INSTANCES  # pylint: disable=line-too-long
```
tomchop, line 155: You can use `pylint` statements at the class or function level, which might make it easier to read

tomchop, line 155: Have you tried assigning `mock_ec2_api.return_value.describe_instances` to its own variable? Might make the code more readable if you reuse it below, but I don't know how that will affect the mock behavior

tests/unittest/aws_test.py
```diff
@@ -0,0 +1,379 @@
...
+    self.assertRaises(ValueError, FAKE_AWS_ACCOUNT.GetVolumesByNameOrId)
+    self.assertRaises(
+        ValueError,
+        FAKE_AWS_ACCOUNT.GetVolumesByNameOrId,
+        volume_id=FAKE_VOLUME.volume_id,
+        volume_name=FAKE_BOOT_VOLUME.name)
+
+  @mock.patch('libcloudforensics.aws.AWSAccount.ClientApi')
+  def testCreateDiskFromSnapshot(self, mock_ec2_api):
+    """Test the creation of a volume from a snapshot."""
```
tomchop, line 288: If you're testing the creation of a Volume, should the test function be renamed for consistency?

tests/unittest/aws_test.py
giovannt0: Gotta love Python, this works well. I've also migrated the other mock statements to their own variable regardless of their re-use or not (only for 80+ chars lines), as I would prefer not to disable `pylint` at class/method level.

PR #58, opened 2020-05-05
Add option to create SSD from snapshot

### LINKED ISSUE #54: Create Disk Copy allows SSD instead of standard disk

giovannt0:
Fixes #54 

### DIFF
--- a/libcloudforensics/gcp.py
+++ b/libcloudforensics/gcp.py
@@ -301,7 +301,8 @@ class GoogleCloudProject:
       disk_name_prefix (str): Optional. String to prefix the disk name with.
       disk_type (str): Optional. URL of the disk type resource describing
           which disk type to use to create the disk. Default is pd-standard. Use
-          pd-ssd to have a SSD disk.
+          pd-ssd to have a SSD disk. You can list all available disk types by
+          running the following command: gcloud compute disk-types list
 
     Returns:
       GoogleComputeDisk: Google Compute Disk.
--- a/libcloudforensics/gcp.py
+++ b/libcloudforensics/gcp.py
@@ -288,14 +288,20 @@ class GoogleCloudProject:
       raise RuntimeError(error_msg)
     return disk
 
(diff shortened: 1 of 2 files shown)


tomchop:
This looks good, but I think the original ask was to also have the Analysis VM be created with an SSD boot disk :) Might be a good thing to add to this PR

giovannt0:
@tomchop ah, yeah that makes sense :p 
I updated my commit to include that as well.

giovannt0:
#60 should be merged before this one.

libcloudforensics/gcp.py
```diff
@@ -296,14 +296,20 @@ def GetDisk(self, disk_name, zone=None):
...
+                             disk_name=None,
+                             disk_name_prefix='',
+                             disk_type='pd-standard'):
     """Create a new disk based on a Snapshot.
 
     Args:
       snapshot (GoogleComputeSnapshot): Snapshot to use.
       disk_name (str): Optional. String to use as new disk name.
       disk_name_prefix (str): Optional. String to prefix the disk name with.
+      disk_type (str): Optional. URL of the disk type resource describing
```
tomchop, line 310: Add something about running `gcloud compute disk-types list` to get a full list of available types?

tests/e2e/gcp_e2e.py
```diff
@@ -158,7 +159,7 @@ def test_end_to_end_other_disk(self):
     # Get the analysis VM and attach the evidence disk to forensic
     self.analysis_vm, _ = gcp.StartAnalysisVm(
         project=self.project_id, vm_name=self.analysis_vm_name, zone=self.zone,
-        boot_disk_size=10, cpu_cores=4,
+        boot_disk_size=10, boot_disk_type='pd-standard', cpu_cores=4,
```
tomchop, line 162: That's the default value - it would be nice to change this test to use a SSD drive and then add an assertion that confirms that the type of the disk that was just created is `pd-ssd`.

PR #64, opened 2020-05-07
Add AWS VM bootstrapping capabilities

giovannt0:
Add capabilities to bootstrap an analysis VM in AWS and attach cloned disks to it, as with GCP.

#58 should be merged before.

### DIFF
--- a/libcloudforensics/aws.py
+++ b/libcloudforensics/aws.py
@@ -22,6 +22,7 @@ import binascii
 import datetime
 import json
 import logging
+import os
 import re
 
 import boto3
@@ -32,7 +33,10 @@ log = logging.getLogger()
 EC2_SERVICE = 'ec2'
 ACCOUNT_SERVICE = 'sts'
 KMS_SERVICE = 'kms'
+# Default Amazon Machine Image to use for bootstrapping instances
+UBUNTU_1804_AMI = 'ami-0013b3aa57f8a4331'
 REGEX_TAG_VALUE = re.compile('^.{1,255}$')
+STARTUP_SCRIPT = 'scripts/startup.sh'
 
 
 class AWSAccount:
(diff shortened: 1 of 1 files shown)


libcloudforensics/aws.py
```diff
@@ -30,7 +31,10 @@
 
 EC2_SERVICE = 'ec2'
 ACCOUNT_SERVICE = 'sts'
+# Default Amazon Machine Image to use for bootstrapping instances
+UBUNTU_1804_AMI = 'ami-0013b3aa57f8a4331'
```
tomchop, line 35: Where does this come from? Is it a universally accessible repo?

libcloudforensics/aws.py
```diff
@@ -526,6 +608,92 @@ def GetAccountInformation(self, info):
...
+    Raises:
+      ValueError: If the requested amount of cores is unavailable.
+    """
+
+    cpus = [1, 2, 4, 8, 16, 32, 40, 48, 64, 96, 128]
+    instance_names = ['t2.small', 'm4.large', 'm4.xlarge', 'm4.2xlarge',
+                      'm4.4xlarge', 'm5.8xlarge', 'm4.10xlarge', 'm5.12xlarge',
+                      'm4.16xlarge', 'm5.24xlarge', 'x1.32xlarge']
+    try:
+      idx = cpus.index(cpu_cores)
```
tomchop, line 659: Maybe replace this by

```python
if cpu_cores not in cpus:
  raise ValueError('Cannot start...')
```

In fact, if there's a 1:1 mapping between "number of cores" and "instance type", isn't it easier to just use a dict for this?

```python
cpu_instance_map = {
    1: 't2.small',
    ...
}
if cpu_cores not in instance_type:
  raise ValueError()
return cpu_instance_map[cpu_cores]
```

libcloudforensics/aws.py
```diff
@@ -854,12 +1065,56 @@ def CreateVolumeCopy(zone,
...
+def StartAnalysisVm(vm_name,
+                    default_availability_zone,
+                    boot_volume_size,
+                    cpu_cores=4,
+                    ami=UBUNTU_1804_AMI,
+                    attach_volume=None,
+                    device_name=None,
+                    dst_account=None):
+  """Start a virtual machine for analysis purposes.
+
```
tomchop, line 1083: Please add a sentence or two explaining what happens if the VM already exists, if the VM exists but is turned off (is that a thing in AWS?) or if the VM doesn't exist.

libcloudforensics/aws.py
```diff
@@ -854,12 +1065,56 @@ def CreateVolumeCopy(zone,
...
+    attach_volume (AWSVolume): Optional. The volume to attach.
+    device_name (str): Optional. The name of the device (e.g. /dev/sdf) for the
+        volume to be attached. Mandatory if attach_volume is provided.
+    dst_account (str): Optional. The AWS account in which to create the
+        analysis VM. This is the profile name that is defined in your AWS
+        credentials file.
+
+  Returns:
+    tuple(AWSInstance, bool): a tuple with a virtual machine object
+    and a boolean indicating if the virtual machine was created or not.
```
tomchop, line 1102: Add indentation here

libcloudforensics/aws.py
giovannt0: This is a public AMI ID that is available on AWS marketplace. Should be available to anyone.

libcloudforensics/aws.py
giovannt0: Ah yeah, brain fart! Good catch, thanks

giovannt0:
#70 should be merged before for ease of rebasing.

PR #65, opened 2020-05-07
Add unittests for AWS VM bootstrapping capabilities

giovannt0:
#64 should be merged before so that tests pass.

### DIFF
--- a/tests/unittest/aws_test.py
+++ b/tests/unittest/aws_test.py
@@ -137,6 +137,22 @@ MOCK_CALLER_IDENTITY = {
     'UserId': 'fake-user-id'
 }
 
+MOCK_DESCRIBE_AMI = {
+    'Images': [{
+        'BlockDeviceMappings': [{
+            'Ebs': {
+                'VolumeSize': None
+            }
+        }]
+    }]
+}
+
+MOCK_RUN_INSTANCES = {
+    'Instances': [{
+        'InstanceId': 'new-instance-id'
+    }]
+}
+
 
 class AWSAccountTest(unittest.TestCase):
   """Test AWSAccount class."""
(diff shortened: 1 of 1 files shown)


tests/unittest/aws_test.py
```diff
@@ -322,16 +338,47 @@ def testCreateVolumeFromSnapshot(self, mock_ec2_api):
...
+
+    # GetOrCreateAnalysisVm(non_existing_vm, boot_volume_size, AMI, cpu_cores).
+    # We mock the GetInstanceById() call to throw a RuntimeError to mimic
+    # an instance that wasn't found.
+    mock_get_instance.side_effect = RuntimeError()
+    vm, created = FAKE_AWS_ACCOUNT.GetOrCreateAnalysisVm(
+        'non-existent-instance-name', 1, 'ami-id', 2)
+    self.assertIsInstance(vm, aws.AWSInstance)
+    self.assertEqual('non-existent-instance-name', vm.name)
+    self.assertTrue(created)
```
tomchop, line 370: The behavior this is testing is identical to the behavior in the previous block. Which actually kinda matches what I expect - GetOrCreateAnalysisVm will create the analysis VM if it doesn't exist.

Isn't there a "create" function you can mock and assert it's actually being called in this case and not in the previous one? (see my comment below)

tests/unittest/aws_test.py
```diff
@@ -340,6 +387,26 @@ def testGenerateVolumeName(self, mock_ec2_api):
     self.assertEqual('prefix-fake-snapshot-d69d57c3-copy', volume_name)
     # pylint: enable=protected-access
 
+  @mock.patch('libcloudforensics.aws.AWSAccount.ClientApi')
+  def testGetBootVolumeConfigByAmi(self, mock_ec2_api):
+    """Test that the boot volume configuration is correctly created."""
+    # pylint: disable=protected-access
```
tomchop, line 393: Where's the protected access you're disabling?

tests/unittest/aws_test.py
```diff
@@ -383,5 +450,91 @@ def testSnapshot(self, mock_ec2_api):
     self.assertTrue(snapshot.name.startswith('my-snapshot'))
 
 
+class AWSTest(unittest.TestCase):
+  """Test the aws.py public methods."""
+
+  @mock.patch('libcloudforensics.aws.AWSVolume.Snapshot')
+  @mock.patch('libcloudforensics.aws.AWSAccount.GetVolumeById')
+  @mock.patch('libcloudforensics.aws.AWSAccount.GetAccountInformation')
+  @mock.patch('libcloudforensics.aws.AWSAccount.ClientApi')
```
tomchop, line 459: You can make use of the `assert_called_*` functions on the mock objects: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called

This allows you to make sure that the functions you mock are being called with the expected arguments.

As a general comment I think that other tests could also benefit from these (especially in `testGetOrCreateAnalysisVm`)

tests/unittest/aws_test.py
giovannt0: I updated that with an additional assertion on `mock_ec2_api.return_value.run_instances.assert_not_called()` in the first block, and `mock_ec2_api.return_value.run_instances.assert_called()` in the second block. The `run_instances` call is the boto3 method to create instances.

tests/unittest/aws_test.py
giovannt0: I added a few more statements on these; and opened #73 to keep track.

PR #66, opened 2020-05-07
Add e2e tests for AWS VM bootstrapping capabilities

giovannt0:
#65 should be merged before so that the tests pass.

### DIFF
--- a/tests/e2e/aws_e2e.py
+++ b/tests/e2e/aws_e2e.py
@@ -59,6 +59,9 @@ class EndToEndTest(unittest.TestCase):
     cls.zone = project_info['zone']
     cls.volume_to_forensic = project_info.get('volume_id', None)
     cls.aws = aws.AWSAccount(cls.zone)
+    cls.analysis_vm_name = 'new-vm-for-analysis'
+    cls.analysis_vm, _ = aws.StartAnalysisVm(
+        cls.analysis_vm_name, cls.zone, 10, 4)
     cls.volumes = []
 
   def test_end_to_end_boot_volume(self):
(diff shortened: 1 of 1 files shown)


tests/e2e/aws_e2e.py
```diff
@@ -83,6 +86,27 @@ def test_end_to_end_boot_volume(self):
...
+    # The forensic instance should be live in the analysis AWS account and
+    # the volume should be attached
+    instance = self.aws.ResourceApi(EC2_SERVICE).Instance(
+        self.analysis_vm.instance_id)
+    self.assertEqual(instance.instance_id, self.analysis_vm.instance_id)
+    for volume in instance.volumes.all():
+      if volume.volume_id == boot_volume_copy.volume_id:
+        return
+    self.fail('Error: could not find the volume {0:s} in instance {1:s}'.format(
+        boot_volume_copy.volume_id, self.analysis_vm_name))
```
tomchop, line 108: If something in the VM Creation logic fails, then your snapshotting tests will fail, making the issue harder to locate.

I would just add an extra test instead of adding this block to both tests.

tests/e2e/aws_e2e.py
```diff
@@ -106,9 +109,48 @@ def test_end_to_end_other_volume(self):
...
+    )
+
+    # The forensic instance should be live in the analysis AWS account and
+    # the volume should be attached
+    instance = self.aws.ResourceApi(EC2_SERVICE).Instance(
+        self.analysis_vm.instance_id)
+    self.assertEqual(instance.instance_id, self.analysis_vm.instance_id)
+    for volume in instance.volumes.all():
+      if volume.volume_id == volume_to_attach.volume_id:
+        return
```
tomchop, line 140: What you want here is `assertIn()` https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn

tests/e2e/aws_e2e.py
```diff
@@ -106,9 +109,45 @@ def test_end_to_end_other_volume(self):
...
+        device_name='/dev/sdp'
+    )
+
+    # The forensic instance should be live in the analysis AWS account and
+    # the volume should be attached
+    instance = self.aws.ResourceApi(EC2_SERVICE).Instance(
+        self.analysis_vm.instance_id)
+    self.assertEqual(instance.instance_id, self.analysis_vm.instance_id)
+    self.assertIn(volume_to_attach.volume_id, list(map(
+        lambda vol: vol.volume_id, instance.volumes.all())))
```
tomchop, line 139: A list comprehension would probably be more readable than `list(map())`

```python
[vol.volume_id for vol in instance volumes.all()]
```

PR #68, opened 2020-05-08
Add run_tests.sh script and fix yapf config.

hacktobeer:
- Add small bash script to execute unit tests.
- Change contribute doc to reflect test script change
- Fix yapf configuration file. 'chromium' style has changed to 'yapf'. See https://github.com/google/yapf/blob/master/CHANGELOG#L17

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -3,7 +3,7 @@
 #     yapf -i -r --style .style.yapf ./turbinia/
 #
 [style]
-based_on_style = chromium
+based_on_style = yapf
 COALESCE_BRACKETS = True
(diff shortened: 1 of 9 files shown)


tomchop:
We're using `nose` as our test runner, so you can just run `nosetests tests` and it will discover and run all tests under the `/tests` directory

hacktobeer:
Would it still be useful for people who want to contribute and easily want to run tests?

hacktobeer:
Removed bash script and adjusted contributing doc to reflect the usage of 'nose' for executing tests.

PR #70, opened 2020-05-08
Add support for sharing encrypted EBS resources cross account

### LINKED ISSUE #63: Deal with EBS volume encryption in AWS

giovannt0:
Few changes:

- Added `encrypted` attribute to `AWSVolume` class.
- When calling `CreateVolumeCopy()` with a destination account that is not the same as the source account, we now check if the target volume is encrypted. If so, a one-time use KMS shared key (between source and destination account) is generated. Subsequently, the initial target volume is duplicated + encrypted with the temporary shared key. We then use this temporary volume for the remainder of the code (i.e., getting a new snapshot out of that temporary volume, sharing it with the destination account, and creating a copy within the destination account).

After execution, temporary volumes, snapshots and keys are deleted. The copy of the volume that sits in the destination account is encrypted with the default EBS/KMS key particular to that account.

Closes #63 

### DIFF
--- a/libcloudforensics/aws.py
+++ b/libcloudforensics/aws.py
@@ -426,16 +426,15 @@ class AWSAccount:
                        '{1:s}'.format(volume_name, REGEX_TAG_VALUE.pattern))
 
     client = self.ClientApi(EC2_SERVICE)
+    create_volume_args = {
+        'AvailabilityZone': snapshot.availability_zone,
+        'SnapshotId': snapshot.snapshot_id,
+        'TagSpecifications': [GetTagForResourceType('volume', volume_name)]
+    }
+    if kms_key_id:
+      create_volume_args['Encrypted'] = True
+      create_volume_args['KmsKeyId'] = kms_key_id
     try:
-      create_volume_args = {
-          'AvailabilityZone': snapshot.availability_zone,
-          'SnapshotId': snapshot.snapshot_id,
-          'TagSpecifications': [GetTagForResourceType('volume', volume_name)]
-      }
-      if kms_key_id:
-        create_volume_args['Encrypted'] = True
-        create_volume_args['KmsKeyId'] = kms_key_id
-
       volume = client.create_volume(**create_volume_args)
       volume_id = volume['VolumeId']
       zone = volume['AvailabilityZone']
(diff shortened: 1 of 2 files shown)


libcloudforensics/aws.py
```diff
@@ -420,12 +427,19 @@ def CreateVolumeFromSnapshot(self,
...
-          TagSpecifications=[GetTagForResourceType('volume', volume_name)])
+      create_volume_args = {
+          'AvailabilityZone': snapshot.availability_zone,
+          'SnapshotId': snapshot.snapshot_id,
+          'TagSpecifications': [GetTagForResourceType('volume', volume_name)]
+      }
+      if kms_key_id:
+        create_volume_args['Encrypted'] = True
+        create_volume_args['KmsKeyId'] = kms_key_id
+
```
tomchop, line 438: You can (probably) move this code out of the try block? Unless `GetTagForResource` type can also raise `ClientError` or `WaiterError`

libcloudforensics/aws.py
```diff
@@ -465,6 +480,72 @@ def GetAccountInformation(self, info):
       raise KeyError('Key must be one of ["UserId", "Account", "Arn"]')
     return account_information.get(info)
 
+  def CreateSharedKMSKey(self, aws_account_id):
```
tomchop, line 483: I would tend to separate the logic in Creation and sharing of the key. Maybe make the second parameter optional? Or split this into two functions (one for creating and one for sharing?)

libcloudforensics/aws.py
```diff
@@ -818,11 +934,28 @@ def CreateVolumeCopy(zone,
         'Account')
 
     if source_account_id != destination_account_id:
+      if volume_to_copy.encrypted:
+        # Generate one-time use KMS key that will be shared with the
+        # destination account.
+        kms_key_id = source_account.CreateKMSKey()
```
tomchop, line 940: These two `CreateKMSKey` and `ShareKMSKeyWithAWSAccount` can also raise `RuntimeError`, so please udpate the docstring accordingly. It might be a good idea to raise more specific errors. Opened #71 to track this.

PR #89, opened 2020-05-14
Add AWS log basics and AWS CLI tools

hacktobeer:
Functionality added:
- AWS CLI tool that has both log querying and volume copy functions
- Add basic AWS CloudTrail querying with filters.


hacktobeer:
I need to commit NextToken support, revoking PR for now. Will get to it tomorrow. Ignore.

PR #90, opened 2020-05-15
Add aws log basics and AWS CLI tool

hacktobeer:
Add AWS log querying functionality
- Query CloudTrail logs using a user supplied filter
- Add AWS CLI tool supporting log querying and (already existing) disk copying


examples/aws_cli.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def LookupLogEvents(args):
+  """Lookup AWS CloudTrail log events.
+
+  Args:
+    args (dict): Arguments from ArgumentParser.
+  """
+  ct = aws.AWSCloudTrail(
+      aws.AWSAccount(default_availability_zone='eu-central-1a'))
```
giovannt0, line 48: Shouldn't this be `default_availability_zone=args.zone` ?

examples/aws_cli.py
```diff
@@ -0,0 +1,104 @@
...
+      print(event)
+
+
+if __name__ == '__main__':
+  parser = argparse.ArgumentParser(description='Demo CLI tool for AWS')
+  subparsers = parser.add_subparsers()
+
+  parser_querylogs = subparsers.add_parser(
+      'querylog', help='Query AWS CloudTrail logs')
+  parser_querylogs.add_argument('--filter', help='Query filter: \'value,key\'')
```
giovannt0, line 72: This will accept only one filter, correct? Does the AWS API supports a list of filters or is it limited to single ones?

examples/aws_cli.py
```diff
@@ -0,0 +1,104 @@
...
+  parser_volumecopy.add_argument(
+      '--volume_id',
+      help='The AWS unique volume ID of the volume to copy. If none '
+      'specified, then --instance_id must be specified and the boot '
+      'volume of the AWS instance will be copied.')
+  parser_volumecopy.add_argument(
+      '--instance_id', help='The AWS unique instance ID')
+  parser_volumecopy.add_argument(
+      '--src_account', help='The name of the profile for the '
+      'source account, as defined in '
```
giovannt0, line 95: Does this+next line fit in one?

examples/aws_cli.py
```diff
@@ -0,0 +1,104 @@
...
+      'the AWS credentials file.')
+  parser_volumecopy.add_argument(
+      '--dst_account', help='The name of the profile for the '
+      'destination account, as defined '
```
giovannt0, line 99: Does this+next line fit in one?

libcloudforensics/aws.py
```diff
@@ -1065,6 +1065,71 @@ def ShareWithAWSAccount(self, aws_account_id):
...
+  def __init__(self, aws_account):
+    """Initialize an AWS CloudTrail client.
+
+    Args:
+      aws_account (AWSAccount): The AWS account to use.
+    """
+
+    self.aws_account = aws_account
+
+  def LookupEvents(self, qfilter=(),
```
giovannt0, line 1084: Add a new line before `qfilter()`

libcloudforensics/aws.py
```diff
@@ -1065,6 +1065,71 @@ def ShareWithAWSAccount(self, aws_account_id):
...
+      # pylint: disable=line-too-long
+      # qfilter = [{'AttributeKey':'Username', 'AttributeValue':'root'}]
+      # starttime = datetime(2020,5,5,17,33,00)
+      # LookupEvents(qfilter=qfilter, starttime=starttime)
+      # Check documentation for qfilter details
+      # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudtrail.html#CloudTrail.Client.lookup_events
+
+    Args:
+      qfilter (tuple): Optional. Filter for the query including 1 AttributeKey and AttributeValue.
+      starttime (datetime): Optional. Start datetime to add to query filter
```
giovannt0, line 1099: Missing punctuation on this+next line (`filter.`) and in returns block (`events.`)

libcloudforensics/aws.py
```diff
@@ -1065,6 +1065,71 @@ def ShareWithAWSAccount(self, aws_account_id):
...
+      endtime (datetime): Optional. End datetime to add to query filter
+
+    Returns:
+      list(dict): A list of events
+    """
+
+    next_token = ''
+    events = []
+
+    client = self.aws_account.ClientApi('cloudtrail')
```
giovannt0, line 1109: Move `'cloudtrail'` to its own constant at the top of the file along with the EC2/KMS constants that are defined

examples/aws_cli.py
hacktobeer: Only one set is supported.
"LookupAttributes (list) --
Contains a list of lookup attributes. Currently the list can contain only one item."
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudtrail.html#CloudTrail.Client.lookup_events

libcloudforensics/aws.py
hacktobeer: Sharp :)

examples/aws_cli.py
giovannt0: Cool, thanks for the clarification :)

examples/aws_cli.py
```diff
@@ -0,0 +1,105 @@
...
+
+  parser_querylogs = subparsers.add_parser(
+      'querylog', help='Query AWS CloudTrail logs')
+  parser_querylogs.add_argument(
+      'zone',
+      help='The AWS zone in which resources are located, e.g. us-east-2b')
+  parser_querylogs.add_argument('--filter', help='Query filter: \'value,key\'')
+  parser_querylogs.add_argument('--start',
+                                help='Start date for query (2020-05-01 11:13:00)',
+                                default='1800-01-01 00:00:00')
```
tomchop, line 78: I would set the default values as None here and deal with that case in the code (see my other comment about default aprameters)

libcloudforensics/aws.py
```diff
@@ -1065,6 +1066,72 @@ def ShareWithAWSAccount(self, aws_account_id):
...
+      list(dict): A list of events.
+    """
+
+    next_token = ''
+    events = []
+
+    client = self.aws_account.ClientApi(CLOUDTRAIL_SERVICE)
+
+    response = client.lookup_events(LookupAttributes=qfilter,
+                                    StartTime=starttime,
```
tomchop, line 1114: What's the behavior of the lib if when either of these times isn't provided?

hacktobeer: It will error out (eg if set to None). That is why I want them to be from 1800->now() if not given.

libcloudforensics/aws.py
```diff
@@ -1065,6 +1066,72 @@ def ShareWithAWSAccount(self, aws_account_id):
...
+                   endtime=datetime.now()):
+    """Lookup events in the CloudTrail logs of this account.
+
+    Example usage:
+      # pylint: disable=line-too-long
+      # qfilter = [{'AttributeKey':'Username', 'AttributeValue':'root'}]
+      # starttime = datetime(2020,5,5,17,33,00)
+      # LookupEvents(qfilter=qfilter, starttime=starttime)
+      # Check documentation for qfilter details
+      # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudtrail.html#CloudTrail.Client.lookup_events
```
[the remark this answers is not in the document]
hacktobeer: it is, directive on line 1092 so I don't get long line errors for the whole comment block.

libcloudforensics/aws.py
```diff
@@ -1065,6 +1066,72 @@ def ShareWithAWSAccount(self, aws_account_id):
...
+    Args:
+      aws_account (AWSAccount): The AWS account to use.
+    """
+
+    self.aws_account = aws_account
+
+  def LookupEvents(self,
+                   qfilter=(),
+                   starttime=datetime(1800, 1, 1),
+                   endtime=datetime.now()):
```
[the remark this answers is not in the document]
hacktobeer: The cool thing about *not* setting them to None is that if you use the function in your IDE it pops up the definition and that would immediately show what value start and end would get if you do not pass in any value. By setting these to None it's unclear and you wold need to add it (at least) to the function documentation.

examples/aws_cli.py
hacktobeer: See my other comment on default parameters.

libcloudforensics/aws.py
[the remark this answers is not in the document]
tomchop: Yes, and the default values should still be added to the function docstring. 

The deeper problem with this is that the `now()` will be evaluated when the whole module loads, not when `LookUp` events is called. All calls to LookUp events use this default value for endtime (the time when the module was loaded), no matter when they are executed in the code.

libcloudforensics/aws.py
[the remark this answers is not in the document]
tomchop: ACk

libcloudforensics/aws.py
tomchop: According to [this](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudtrail.html#CloudTrail.Client.lookup_events), all parameters are optional, so only passing the args if they're specified sounds like a much better solution here. 

Something like:

```python
params = {'LookupAttributes': qfilter}
if starttime:
  params['StarTime'] = starttime
if endtime:
  params['EndTime'] = endtime

client.lookup_events(**params)
```

libcloudforensics/aws.py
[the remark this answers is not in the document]
hacktobeer: ack, done.

libcloudforensics/aws.py
hacktobeer: that's way more logical, changes made.

hacktobeer:
Processed comments, rewrote to use **param to make it more flexible and clean, thanks for the comments :).

examples/aws_cli.py
hacktobeer: done

PR #91, opened 2020-05-15
Add YAPF single parameter per line configuration.

### LINKED ISSUE #84: Make YAPF use one parameter per line
YAPF should put every function parameter on a separate line for readability. 
Add below to YAPF style file and create a separate re-style PR for all code.

SPLIT_ALL_COMMA_SEPARATED_VALUES = True

hacktobeer:
If a comma separated list (dict, list, tuple, or function def) is on a line that is too long, split such that all elements are on a single line.

Style preparation for refactor. Closes #84


.style.yapf
```diff
@@ -5,7 +5,8 @@
 [style]
 based_on_style = yapf
 COALESCE_BRACKETS = True
+SPLIT_ALL_COMMA_SEPARATED_VALUES = True
```
tomchop, line 8: I think we can use `split_all_top_level_comma_separated_values` instead to achieve a reasonable result: https://github.com/google/yapf/issues/574

hacktobeer: we could, as long as we are okay with having to possible do manual cleanup when embedded args/structs get too long or messy?

tomchop: That's fine by me.

PR #92, opened 2020-05-15
Create images using disks from another project

sa3eed3ed:
Adding Delete Image Function
Force image creation if source disk is in use
Fixing doc Strings
Fix bug while creating GoogleComputeImage

### DIFF
--- a/libcloudforensics/gcp.py
+++ b/libcloudforensics/gcp.py
@@ -1346,18 +1346,13 @@ class GoogleComputeImage(GoogleComputeBaseResource):
 
   def Delete(self):
     """Delete Compute Disk Image from a project.
-
-    Returns:
-      dict: Holding the response of a get operation on an API object of type
-          globalOperations..
     """
 
     gce_image_client = self.project.GceApi().images()
     request = gce_image_client.delete(
         project=self.project.project_id, image=self.name)
     response = request.execute()
-    result = self.project.BlockOperation(response)
-    return result
+    self.project.BlockOperation(response)
 
 
 def CreateDiskCopy(
--- a/libcloudforensics/gcp.py
+++ b/libcloudforensics/gcp.py
@@ -267,7 +267,8 @@ class GoogleCloudProject:
           GlobalZone.
 
     Returns:
-      str: Operation result in JSON format.
+      dict: Holding the response of a get operation on an API object of type
+          zoneOperations or globalOperations.
 
     Raises:
       RuntimeError: If API call failed.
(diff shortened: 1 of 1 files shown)


libcloudforensics/gcp.py
```diff
@@ -1340,6 +1344,21 @@ def ExportImage(self, gcs_output_folder, output_name=None):
     cloud_build.BlockOperation(response)
     log.info('Image {0:s} exported to {1:s}.'.format(self.name, full_path))
 
+  def Delete(self):
+    """Delete Compute Disk Image from a project.
+
+    Returns:
+      dict: Holding the response of a get operation on an API object of type
+          globalOperations..
```
giovannt0, line 1352: duplicate `.`

libcloudforensics/gcp.py
```diff
@@ -1340,6 +1344,21 @@ def ExportImage(self, gcs_output_folder, output_name=None):
...
+    """
+
+    gce_image_client = self.project.GceApi().images()
+    request = gce_image_client.delete(
+        project=self.project.project_id, image=self.name)
+    response = request.execute()
+    result = self.project.BlockOperation(response)
+    return result
```
giovannt0, line 1360: I see we don't return anything for similar methods (e.g. `GoogleComputeSnapshot.Delete()`). Do we need the response here, or is it missing in the first place in the other method(s)?

libcloudforensics/gcp.py
```diff
@@ -723,14 +724,15 @@ def CreateImageFromDisk(self, src_disk, name=None):
...
                 src_disk=src_disk.name)
     }
     gce_image_client = self.GceApi().images()
-    request = gce_image_client.insert(project=self.project_id, body=image_body)
+    request = gce_image_client.insert(
+        project=self.project_id, body=image_body, forceCreate=True)
     response = request.execute()
     self.BlockOperation(response)
-    return GoogleComputeImage(self.project_id, None, name)
+    return GoogleComputeImage(self, None, name)
```
giovannt0, line 735: Why is zone `None` here?

sa3eed3ed: yes, Images are global resources in GCP and zone should be None, in addition any mutating operation will directly create an GlobalOperation object.

libcloudforensics/gcp.py
sa3eed3ed: mmm, since blockOperation will raise errors, we can get rid of this return for the sake of consistency.

PR #93, opened 2020-05-19
Initial refactor proposal

giovannt0:
As discussed. This is mostly architectural (doc-strings and all weren't updated), as a first draft.


examples/gcp_cli.py
```diff
@@ -16,7 +16,9 @@
 
 import argparse
 import json
-from libcloudforensics import gcp
+from libcloudforensics.providers.gcp.internal.project import GoogleCloudProject
```
tomchop, line 19: Please import the module and not the class directly:

```python
from libcloudforensics.providers.gcp.internal import project
project.GoogleCloudProject(...)
```

Alternatively, you can also define in /gcp/internal/__init__.py which modules to expose here, like

```python
from gcp import GoogleCloudProject # You're effectively importing a class and not a module, but it's really clear where it's coming from since it's all contained in the __init__ file, so I guess it's OK
```

So you can do

```python
from libcloudforensics.providers.gcp import internal as gcp_internal
gcp_internal.GoogleCloudProject
```

libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -0,0 +1,155 @@
...
+  if not service_built:
+    error_msg = (
+        'Failures building service {0:s} caused by multiple '
+        'timeouts').format(service_name)
+    raise RuntimeError(error_msg)
+
+  return service
+
+
+def ReadStartupScript():
```
tomchop, line 127: Isn't this the same as in aws.py?

examples/gcp_cli.py
giovannt0: I opt for the second approach, given the amount of submodules we have under `internal` I find it cleaner.

libcloudforensics/providers/gcp/internal/common.py
giovannt0: Ah, yes. I moved that to its own module so both can call the same method.

giovannt0:
@tomchop As discussed, PR comments commits dropped for now; we can cherry-pick them later in smaller PRs.

libcloudforensics/providers/gcp/internal/project.py
```diff
@@ -0,0 +1,681 @@
...
+      project_id (str): The name of the project.
+      default_zone (str): Optional. Default zone to create new resources in. N
+          one means GlobalZone.
+    """
+
+    self.project_id = project_id
+    self.default_zone = default_zone
+    self.gce_api_client = None
+
+  def GceApi(self):
```
sa3eed3ed, line 63: should we split this file per API services and move each service to a separate file, similar to log.py, something like gce.py, gcf.py, gcb.py and then the user can optionally initialize an object created by each of these modules inside the project. Later the caller will do something like project.compute.create_instance. This will make porting new functions easy since it is structured in a way similar to the GCP APIs.

PR #96, opened 2020-05-19
Switch import from classes to modules

giovannt0:
Note that pylint fails atm with a cyclic import; this will be fixed in another PR when we change the `forensics` module(s)


giovannt0:
Merging, pending following changes:

- Migrate `ReadStartupScript` so that all providers call the same method
- Remove `forensics_interface`

PR #113, opened 2020-06-03
Add Paginate method for GCP

### LINKED ISSUE #80: Migrate loops body to separate method
`while not have_all_tokens:` kinda loops, across gcp.py and aws.py

giovannt0:
Add the second part of the fixit for #80, following #110 
Closes #80 

### DIFF
--- a/libcloudforensics/providers/gcp/internal/common.py
+++ b/libcloudforensics/providers/gcp/internal/common.py
@@ -125,7 +125,7 @@ def CreateService(service_name, api_version):
 
 
 class GoogleCloudComputeClient:
-  """Class represnting Google Cloud Compute API client.
+  """Class representing Google Cloud Compute API client.
 
   Attributes:
     project_id (str): Project name.
@@ -136,7 +136,7 @@ class GoogleCloudComputeClient:
     """Initialize Google Cloud Engine API client object.
 
     Args:
-     project_id (str): Optional. Project name. Nedded to use BlockOperation.
+     project_id (str): Optional. Project name. Needed to use BlockOperation.
     """
     self._gce_api_client = None
     self.project_id = project_id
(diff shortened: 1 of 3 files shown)


libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -187,3 +187,45 @@ def BlockOperation(self, response, zone=None):
...
+      error_msg = (
+          '{0:s}\n'
+          'Something is wrong with your Application Default '
+          'Credentials. Try running: '
+          '$ gcloud auth application-default login'.format(str(exception)))
+      raise RuntimeError(error_msg)
+    responses.append(response)
+    next_token = response.get('nextPageToken')
+    if not next_token:
+      break
```
tomchop, line 230: `return responses` here?

libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -187,3 +187,45 @@ def BlockOperation(self, response, zone=None):
...
+  responses = []
+  next_token = None
+  while True:
+    if throttle:
+      time.sleep(1)
+    if next_token:
+      kwargs['pageToken'] = next_token
+    try:
+      request = getattr(client, func)
+      response = request(**kwargs).execute()
```
tomchop, line 219: What is `response` in this case? Can it be a list? if so, you might want to extend instead of append in L227, otherwise you'll end up with a list of lists.

libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -187,3 +187,45 @@ def BlockOperation(self, response, zone=None):
       if result['status'] == 'DONE':
         return result
       time.sleep(5)  # Seconds between requests
+
+
+def ExecuteAndPaginate(client, func, kwargs, throttle=False):
```
tomchop, line 192: Why is this called `ExecuteAndPaginate`? Even if it indeed uses pagination under the hood, does it really change anything for the caller? It doesn't seem like the caller can specify page sizes or a given page number.

libcloudforensics/providers/gcp/internal/common.py
giovannt0: My understanding of https://cloud.google.com/compute/docs/reference/rest/v1/ and similar APIs is that it always returns a JSON body as a response, never a list of JSON bodies, so we shouldn't encounter the case you're mentioning.

libcloudforensics/providers/gcp/internal/common.py
giovannt0: I have changed it to `ExecuteRequest`

libcloudforensics/providers/gcp/internal/common.py
tomchop: OK cool. The `list` I was anticipating is actually in the `'items`' key of the response; that's where the pagination occurs. Thanks!

PR #117, opened 2020-06-04
Refactor attach_disk / attach_volume to list of volume names / ids

### LINKED ISSUE #88: Refactor attach_volume in StartAnalysisVm() methods
The idea is to pass a list of strings to `StartAnalysisVm()` for the `attach_volume` / `attach_disk` parameter instead of an `AWSVolume` or a `GoogleComputeDisk` object. The list of strings would ideally be a list of volume IDs/names to attach. This will make `StartAnalysisVm()` more flexible (e.g. if wanting to use these methods in the CLI tool without defining volumes/disks in the first place)

giovannt0:
Closes #88 

### DIFF
--- a/libcloudforensics/providers/gcp/forensics.py
+++ b/libcloudforensics/providers/gcp/forensics.py
@@ -114,7 +114,7 @@ def StartAnalysisVm(project,
         which disk type to use to create the disk. Use pd-standard for a
         standard disk and pd-ssd for a SSD disk.
     cpu_cores (int): The number of CPU cores to create the machine with.
-    attach_disks (list(str)): Optional. List of disk names to attach.
+    attach_disks (list[str]): Optional. List of disk names to attach.
     image_project (str): Optional. Name of the project where the analysis VM
         image is hosted.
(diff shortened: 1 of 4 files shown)


libcloudforensics/providers/gcp/forensics.py
```diff
@@ -114,7 +114,7 @@ def StartAnalysisVm(project,
         which disk type to use to create the disk. Use pd-standard for a
         standard disk and pd-ssd for a SSD disk.
     cpu_cores (int): The number of CPU cores to create the machine with.
-    attach_disk (list(GoogleComputeDisk)): Optional. List of disks to attach.
+    attach_disks (list(str)): Optional. List of disk names to attach.
```
tomchop, line 117:
```suggestion
    attach_disks (list[str]): Optional. List of disk names to attach.
```

I think that's what the log2timeline convention is.

giovannt0: Changed for this PR; opened #128 so we can enforce this across the codebase in a separate PR.

PR #119, opened 2020-06-04
Add option to copy volume in a different destination region in AWS

### LINKED ISSUE #86: Add destination zone for snapshots
- [x] AWS: #119 
- [x] GCP

Edit 2020-06-04 [giovannt0]: GCP already offers this in `CreateDiskCopy()` with the `zone` parameter. As such, removing GCP label.

giovannt0:
Closes #86 

E.g usage: 

```
python -m examples.libcloudforensics aws us-west-1a copydisk --dst_zone us-east-2c --volume_id vol-xxx
```

### DIFF
--- a/examples/aws_cli.py
+++ b/examples/aws_cli.py
@@ -62,10 +62,12 @@ def CreateVolumeCopy(args):
     args (dict): Arguments from ArgumentParser.
   """
   print('Starting volume copy...')
-  volume_copy = forensics.CreateVolumeCopy(
-      args.zone, dst_zone=args.dst_zone, instance_id=args.instance_id,
-      volume_id=args.volume_id, src_account=args.src_account,
-      dst_account=args.dst_account)
+  volume_copy = forensics.CreateVolumeCopy(args.zone,
+                                           dst_zone=args.dst_zone,
+                                           instance_id=args.instance_id,
+                                           volume_id=args.volume_id,
+                                           src_account=args.src_account,
+                                           dst_account=args.dst_account)
   print(
       'Done! Volume {0:s} successfully created. You will find it in '
       'your AWS account under the name {1:s}.'.format(
--- a/examples/aws_cli.py
+++ b/examples/aws_cli.py
@@ -63,8 +63,9 @@ def CreateVolumeCopy(args):
   """
   print('Starting volume copy...')
   volume_copy = forensics.CreateVolumeCopy(
-      args.zone, instance_id=args.instance_id, volume_id=args.volume_id,
-      src_account=args.src_account, dst_account=args.dst_account)
+      args.zone, dst_zone=args.dst_zone, instance_id=args.instance_id,
+      volume_id=args.volume_id, src_account=args.src_account,
+      dst_account=args.dst_account)
   print(
       'Done! Volume {0:s} successfully created. You will find it in '
       'your AWS account under the name {1:s}.'.format(
--- a/examples/libcloudforensics.py
+++ b/examples/libcloudforensics.py
@@ -90,6 +90,9 @@ if __name__ == '__main__':
             'List EBS volumes in AWS account.')
   AddParser('aws', aws_subparsers, 'copydisk', 'Create an AWS volume copy.',
             args=[
+                ('--dst_zone', 'The AWS zone in which to copy the volume. By '
+                               'default this is the same as \'zone\'.',
+                 None),
                 ('--instance_id', 'The AWS unique instance ID', None),
                 ('--volume_id', 'The AWS unique volume ID of the volume to '
                                 'copy. If none specified, then --instance_id '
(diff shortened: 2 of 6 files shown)


examples/aws_cli.py
```diff
@@ -63,8 +63,9 @@ def CreateVolumeCopy(args):
   """
   print('Starting volume copy...')
   volume_copy = forensics.CreateVolumeCopy(
-      args.zone, instance_id=args.instance_id, volume_id=args.volume_id,
-      src_account=args.src_account, dst_account=args.dst_account)
+      args.zone, dst_zone=args.dst_zone, instance_id=args.instance_id,
```
tomchop, line 66: One arg per line will make things much readable here IMO

examples/libcloudforensics.py
```diff
@@ -90,6 +90,9 @@ def AddParser(provider, provider_parser, func, func_helper, args=None):
             'List EBS volumes in AWS account.')
   AddParser('aws', aws_subparsers, 'copydisk', 'Create an AWS volume copy.',
             args=[
+                ('--dst_zone', 'The AWS zone in which to copy the volume. By '
+                               'default this is the same as \'zone\'.',
```
tomchop, line 94: FYI - you can use double quotes instead of escaping the single quotes. If you really want single quotes in your text, then it's OK to use double quotes for your string:

```python
"This bug has been 'fixed'"
'This bug has been "fixed"'
```

libcloudforensics/providers/aws/forensics.py
```diff
@@ -62,7 +63,9 @@ def CreateVolumeCopy(zone,
       dst_account='forensics')
 
   Args:
-    zone (str): The zone within the region to create the new resource in.
+    zone (str): The AWS zone in which the volume is located, e.g. us-east-2b.
```
tomchop, line 66:
```suggestion
    zone (str): The AWS zone in which the volume is located, e.g. 'us-east-2b'.
```

libcloudforensics/providers/aws/forensics.py
```diff
@@ -117,12 +120,16 @@ def CreateVolumeCopy(zone,
...
-        # Delete the initial snapshot
-        snapshot.Delete()
-        snapshot = temporary_snapshot
+        snapshot = snapshot.Copy(kms_key_id=kms_key_id, delete=True)
       snapshot.ShareWithAWSAccount(destination_account_id)
 
+    if dst_zone:
+      # Assign the new zone to the destination account and assign it to the
+      # snapshot so that it can copy it
+      destination_account.__init__(dst_zone, aws_profile=dst_account)
```
tomchop, line 129: I are we calling `__init__` here and not through the normal python constructor (`destination_account()`) ?
If this is functionality we need to call to set things up, then please add a function to the destination_account's class to do so.

libcloudforensics/providers/aws/internal/account.py
```diff
@@ -567,7 +567,10 @@ def ShareKMSKeyWithAWSAccount(self, kms_key_id, aws_account_id):
         'Action': [
             'kms:Encrypt',
             'kms:Decrypt',
-            'kms:ReEncrypt*'
+            'kms:ReEncrypt*',
+            # This permission is necessary to transfer encrypted EBS
```
tomchop, line 571: Please name the permission you're referring to explicitly.

libcloudforensics/providers/aws/internal/ebs.py
```diff
@@ -204,17 +231,33 @@ def Copy(self, kms_key_id=None):
...
-          name='{0:s}-copy'.format(self.snapshot_id)
-      )
     except client.exceptions.ClientError as exception:
       raise RuntimeError('Could not copy snapshot {0:s}: {1:s}'.format(
           self.snapshot_id, str(exception)))
 
+    snapshot_copy = AWSSnapshot(
+        # If the call above was successful, the response contains the new
+        # snapshot ID
+        response['SnapshotId'],
```
tomchop, line 241: You're saying "if the call above was successful", but what happens if it isn't? If the answer to that is that it would be caught in the exception on L234, then it's enough to say "The response contains the new snapshot ID"

libcloudforensics/providers/aws/forensics.py
giovannt0: Hmmm I don't see why I did that :D I reverted to default constructor.

libcloudforensics/providers/aws/internal/account.py
giovannt0: Added, also added a small blob for the perms above this one to make it clearer.

libcloudforensics/providers/aws/internal/ebs.py
giovannt0: Ah, this was in the `try` block before hence the comment. I removed it. L234 will indeed catch an exception is something goes bad.

PR #126, opened 2020-06-05
Enhance e2e tests

### LINKED ISSUE #120: Add more e2e tests for AWS
With new capabilities brought by #70 and #119 and seeing how granular AWS IAM permissions can get when it comes to identifying the minimum set of required permissions to use the `CreateVolumeCopy()` functionality in different scenarios, we need more e2e tests that verify these scenarios:

- [x] Boot volume copy
- [x] Specific volume copy
- [x] Bootstrap VM
- [x] Volume copy in a different zone 
- [x] Encrypted volume copy
- [x] Encrypted volume copy in a different zone

giovannt0:
As explained in #120.

Closes #120.


tests/providers/aws/aws_e2e.py
```diff
@@ -132,22 +150,35 @@ def test_end_to_end_vm(self):
...
+    Verify that a volume created by the library is in AWS.
+
+    Args:
+      aws_account (AWSAccount): The AWS account to use.
+      volume_id (str): The volume ID of the volume to verify.
+    """
+    self.volumes.append(
+        (aws_account, aws_account.ResourceApi(EC2_SERVICE).Volume(volume_id)))
+    self.assertEqual(self.volumes[-1][1].volume_id, volume_id)
+
```
tomchop, line 168: I think tests would be clearer if the actual assertion was in the tests, rather than hidden in a function. (It doesn't matter if you repeat yourself in tests - it's better they are explicit than very elegant and convoluted)

I see you're saving modules for cleanup at the end. In that case, it might be better to have a function very explicitly named something like `_StoreVolumeForTestCleanup`. that does the appending. Anyone reading the test will know what it's about and that it's probably safe to ignore.

giovannt0: Thanks for the clarification, changed :)

tests/providers/aws/aws_e2e.py
```diff
@@ -56,99 +52,139 @@ def setUpClass(cls):
...
+        self.zone, volume_id=self.encrypted_volume_to_copy)
+    self._StoreVolumeForCleanup(self.aws, volume_copy.volume_id)
+    self.assertEqual(self.volumes[-1][1].volume_id, volume_copy.volume_id)
+
+  def testEncryptedVolumeCopyToOtherZone(self):
     """End to end test on AWS.
 
-    This tests that an analysis VM is correctly created and that a volume
-        passed to the attach_volume parameter is correctly attached.
+    Test copying a specific encrypted volume to a different AWS availability
```
tomchop, line 125: I'm not sure long docstring descriptions need hanging indents.

tests/providers/aws/aws_e2e.py
```diff
@@ -56,99 +52,139 @@ def setUpClass(cls):
...
+        zone.
```
tomchop, line 126: I'm not sure long docstring descriptions need hanging indents.

tests/providers/aws/aws_e2e.py
```diff
@@ -56,99 +52,139 @@ def setUpClass(cls):
...
+    """
+
+    if not self.volume_to_copy:
+      return
 
-  def test_end_to_end_other_volume(self):
+    volume_copy = forensics.CreateVolumeCopy(
+        self.zone, volume_id=self.volume_to_copy)
+    self._StoreVolumeForCleanup(self.aws, volume_copy.volume_id)
+    self.assertEqual(self.volumes[-1][1].volume_id, volume_copy.volume_id)
```
tomchop, line 91: I'm not sure I understand what's going on here. You're comparing the volume_id attribute of the volume you just appended to itself? `CreateVolumeCopy` could just return a random volume object and the test would still pass.

I think a better e2e test would be to call `CreateVolumeCopy`, then try to confirm that this had the action you wanted in the remote AWS project, either:

* through a unique attribute set server-side by AWS in the response
* or by listing the remove volumes and confirming that the volume ID that you initially received is there

giovannt0: Ah yeah, that doesn't make much sense. I changed that!

PR #129, opened 2020-06-05
Change type declaration in docstrings following l2t style guide

### LINKED ISSUE #128: Enforce l2t style guide for parameter types
Make sure docstrings are on par with https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md

giovannt0:
Closes #128 

### DIFF
--- a/examples/libcloudforensics.py
+++ b/examples/libcloudforensics.py
@@ -50,7 +50,7 @@ def AddParser(provider, provider_parser, func, func_helper, args=None):
     func (str): The name of the function to look for in the given provider
         and to add parsing options for.
     func_helper (str): A helper text describing what the function does.
-    args (list(tuple(str, str, str|None))): Optional. A list of arguments to add
+    args (list[tuple[str, str, str|None]]): Optional. A list of arguments to add
         to the parser. Each argument is a tuple containing the action (str) to
         add to the parser, a helper text (str), and a default value (str,
         optional).
--- a/examples/libcloudforensics.py
+++ b/examples/libcloudforensics.py
@@ -50,7 +50,7 @@ def AddParser(provider, provider_parser, func, func_helper, args=None):
     func (str): The name of the function to look for in the given provider
         and to add parsing options for.
     func_helper (str): A helper text describing what the function does.
-    args (list[tuple[str, str, str|None]]): Optional. A list of arguments to add
+    args (list[tuple]): Optional. A list of arguments to add
         to the parser. Each argument is a tuple containing the action (str) to
         add to the parser, a helper text (str), and a default value (str,
         optional).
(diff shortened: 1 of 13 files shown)


examples/libcloudforensics.py
```diff
@@ -50,7 +50,7 @@ def AddParser(provider, provider_parser, func, func_helper, args=None):
     func (str): The name of the function to look for in the given provider
         and to add parsing options for.
     func_helper (str): A helper text describing what the function does.
-    args (list(tuple(str, str, str|None))): Optional. A list of arguments to add
+    args (list[tuple[str, str, str|None]]): Optional. A list of arguments to add
```
tomchop, line 53: Are you saying that it can be a list of tuples (e.g. `[(a, b, c)]`) *OR* a list of `Nones` (e.g. `[None]`)? If so it might cause problems if we're trying to unpack on iteration:

```python
for action, helper, default in args:
  # do something
```

This will fail if the list contains a non-tuple element.

For more complex structures like this, consider using [NamedTuples](https://docs.python.org/3/library/collections.html#collections.namedtuple)

libcloudforensics/providers/aws/internal/account.py
```diff
@@ -104,13 +104,13 @@ def ListInstances(self, region=None, filters=None, show_terminated=False):
       region (str): Optional. The region from which to list instances.
           If none provided, the default_region associated to the AWSAccount
           object will be used.
-      filters (list(dict)): Optional. Filters for the query.
+      filters (list[dict[str, str|list[str]]]): Optional. Filters for the query.
```
tomchop, line 107: Same, this sounds like it could possible break easily

examples/libcloudforensics.py
giovannt0: I'm trying to say that this is a list of tuples in which each tuple's 3rd element might be a `str` or `None`

libcloudforensics/providers/aws/internal/account.py
giovannt0: According to boto3's doc, filters are just `dict` structures which keys (of type `str`) map directly to another `str` or to a list of `str`. How would you like to have this kind of type declared?

libcloudforensics/providers/aws/internal/common.py
```diff
@@ -36,7 +36,7 @@ def GetTagForResourceType(resource, name):
     name (str): The name of the resource.
 
   Returns:
-    dict: A dictionary for AWS Tag Specifications.
+    dict[str, str|list[dict]]: A dictionary for AWS Tag Specifications.
```
giovannt0, line 39: I left this one as is since the return statement below clearly shows the format

giovannt0:
@tomchop as discussed, made it a bit clearer and added links to documentation for some of the structures.

PR #143, opened 2020-06-11
Fix GCP CLI arguments.

hacktobeer:


hacktobeer:
There are inconsistencies on how we handle optional/mandatory arguments in the the GCP and AWS copy functionality. This needs to be fixed somewhere else. Closing PR for now.

PR #146, opened 2020-06-12
Make AWS tagging more generic

### LINKED ISSUE #145: Implement a more generic tagging feature for AWS
[`GetTagForResourceType()`](https://github.com/google/cloud-forensics-utils/blob/master/libcloudforensics/providers/aws/internal/common.py#L35) is currently leveraged to apply only the `Name` tag to resources. This functionality should be more generic and allow a user to provide (e.g. via custom configuration file) which Keys/Values they want to provide in addition to the `Name` tag

Tags tend to be a critical part of organization in an AWS account and help provide clear context and accountability, among other things.

giovannt0:
As suggested in #145, make tagging in AWS more generic. 

- Rename `GetTagForResourceType` into `CreateTags`.
- `CreateTags` now takes a list of tags to add for the given resource.
- Possibilities to start analysis VMs and create disk copies with a dictionary of tags.

Closes #145.


libcloudforensics/providers/aws/internal/account.py
```diff
@@ -430,7 +430,7 @@ def CreateVolumeFromSnapshot(
         'AvailabilityZone': snapshot.availability_zone,
         'SnapshotId': snapshot.snapshot_id,
         'TagSpecifications':
-            [common.GetTagForResourceType('volume', volume_name)]
+            [common.CreateTags(common.VOLUME, {'Name': volume_name})]
```
sa3eed3ed, line 433: It looks to me that you can add multiple tags during EC2 instance creation, if this is the case wouldn't it make sense to add an optional tags(dict[str,str]) parameter to these functions:
CreateVolumeFromSnapshot, GetOrCreateAnalysisVm, Snapshot

you can make it optional and default to {'Name':name} only if None. This can come to handy if an analyst wants to tag the instance with additional information (ex: ticket number), Turbinia does that while starting a VM in GCP, there it relies on add_label after creating the vm, however I will add labels parameter to all the create_ methods in gcp if you add it here for consistency and to have it done in one step instead of two, the api insert request takes labels dict while creation there as well. If you think this is a good idea, open an issue for GCP and assign it to me.

giovannt0:
Thanks @sa3eed3ed, I think this is a great idea. I've updated the PR, and updated the CLI tool as well. This allows users to call the tool as follow:

```
libcloudforensics aws 'zone' copydisk --volume_id='vol-xxx' --tags='{"customTag": "blah"}'

--

libcloudforensics aws 'zone' copydisk --volume_id='vol-xxx' --tags='{"Name": "some-name", "customTag": "blah"}'
```

In the first case, a volume name will be generated and added as `Name` tag (which is the current behaviour).

In the second case, users can overwrite the name by specifying one on the CLI.

Any other tag added to the dictionary (be it on the CLI or in `CreateVolumeFromSnapshot`, `GetOrCreateAnalysisVm`, `Snapshot` which were all updated) will be added to the AWS resource as well. Note that the `snapshot_name` parameter of the `Snapshot` method was removed as it makes more sense to pass it in the tags directly now.

I've also opened #157 for GCP.

libcloudforensics/providers/aws/internal/ebs.py
```diff
@@ -119,22 +120,24 @@ def Snapshot(self, snapshot_name: Optional[str] = None) -> 'AWSSnapshot':
...
 
+    if not tags:
+      tags = {}
+
+    snapshot_name = tags.get('Name') or self.volume_id
     timestamp = datetime.now().strftime('%Y%m%d%H%M%S')
-    if not snapshot_name:
-      snapshot_name = self.volume_id
     truncate_at = 255 - len(timestamp) - 1
     snapshot_name = '{0}-{1}'.format(snapshot_name[:truncate_at], timestamp)
```
sa3eed3ed, line 129: Is it desired here to add a timestamp suffix to the name in case the name is given by the user as a parameter? or should it only be in case no name is given? I am thinking in a case where the user create a snapshot of the volume with a certain name tag and then search for the same name tag later in the code and find no match, but I am not fully aware how tags work in AWS. If this makes sense to you:

```suggestion
      truncate_at = 255 - len(timestamp) - 1
      snapshot_name = '{0}-{1}'.format(snapshot_name[:truncate_at], timestamp)
```

giovannt0: I think this was initially to be on par with what we had in GCP; however there it is required since the snapshot name is used as ID by GCP (the timestamp was there to limit name collisions). This doesn't matter in AWS since tags are not unique between resources anyways.

I removed it, and opted to go with `volume_id + '-snapshot'` for the default snapshot name if no tags are provided.

libcloudforensics/providers/aws/forensics.py
```diff
@@ -136,8 +139,13 @@ def CreateVolumeCopy(zone: str,
       snapshot.aws_account = destination_account
       snapshot = snapshot.Copy(delete=True, deletion_account=source_account)
 
+    kwargs = {'tags': tags}  # type: Dict[str, Any]
+    if tags and tags.get('Name'):
+      kwargs['volume_name'] = tags['Name']
+    else:
+      kwargs['volume_name_prefix'] = 'evidence'
```
tomchop, line 146: I don't know how I feel about using `kwargs` for a single parameter here (it was a bit confusing to me because I was looking for other parts of the code that modified it too). It might be more clear to call the function in the if blocks directly.

giovannt0: Changed

PR #154, opened 2020-06-18
Add 'body' request field if GCP API needs them.

hacktobeer:
Some google cloud apis expect the request to be in a 'body' field. From the discovery api it is unclear which calls are expecting those. 

This PR
* fixes execute query API call
* takes care of nextpage token

### DIFF
--- a/libcloudforensics/providers/gcp/internal/common.py
+++ b/libcloudforensics/providers/gcp/internal/common.py
@@ -232,7 +232,10 @@ def ExecuteRequest(client: 'googleapiclient.discovery.Resource',
     if throttle:
       time.sleep(1)
     if next_token:
-      kwargs['pageToken'] = next_token
+      if 'body' in kwargs:
+        kwargs['pageToken'] = next_token
+      else:
+         kwargs['pageToken'] = next_token
     try:
       request = getattr(client, func)
       response = request(**kwargs).execute()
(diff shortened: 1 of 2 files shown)


libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -207,7 +207,7 @@ def BlockOperation(self,
 
 def ExecuteRequest(client: 'googleapiclient.discovery.Resource',
                    func: str,
-                   kwargs: Dict[str, str],
+                   kwargs: Union[Dict[str, str], Dict[str, Dict[str, str]]],
```
giovannt0, line 210: To be consistent with the rest of the codebase I would rather type hint such `Dict` as `Dict[str, Any]` (and thus get rid of the `Union` and corresponding import) otherwise it quickly gets lengthy, wdyt?

hacktobeer: sgtm, done.

sa3eed3ed:
Instead of implementing pagination and sending the next_page token inside the ExecuteRequest function, I believe we should change this and use the native list_next and aggregatedList_next functions which are exposed by all paginating GCP API functions
EX:
Logging: http://googleapis.github.io/google-api-python-client/docs/dyn/logging_v2.entries.html
Functions: http://googleapis.github.io/google-api-python-client/docs/dyn/cloudfunctions_v1.operations.html
Compute: https://developers.google.com/resources/api-libraries/documentation/compute/v1/python/latest/compute_v1.instances.html#aggregatedList_next

I can't find any paginating api that doesn't implement {function}_next

This is not really related to this PR, will start a new issue for that.

PR #160, opened 2020-06-22
Expose StartVM functionality through CLI.

### LINKED ISSUE #118: CLI tool should support creating/starting an analysis VM

hacktobeer:
Create and start forensic VMs through the CLI tool.

```
AWS: $ libcloudforensics aws us-east-2a startvm analysis-vm
GCP: $ libcloudforensics gcp project-test startvm analysis-vm us-central-1f
```

Closes #118


examples/cli.py
```diff
@@ -124,6 +124,14 @@ def Main() -> None:
                 ('--start', 'Start date for query (2020-05-01 11:13:00)', None),
                 ('--end', 'End date for query (2020-05-01 11:13:00)', None)
             ])
+  AddParser('aws', aws_subparsers, 'startvm', 'Start a forensic analysis VM.',
+            args=[
+                ('instance_name', 'Name of the EC2 instance to create.',
```
giovannt0, line 129: `to create or re-use`, since we can use existing analysis vms.

examples/cli.py
```diff
@@ -141,6 +149,15 @@ def Main() -> None:
                 ('--disk_name', 'Name of the disk to copy. If None, the boot '
                                 'disk of the instance will be copied.', None)
             ])
+  AddParser('gcp', gcp_subparsers, 'startvm', 'Start a forensic analysis VM.',
+            args=[
+                ('instance_name', 'Name of the GCE instance to create.',
```
giovannt0, line 154: Similar as above

hacktobeer: That's a good idea.

For GCP that's easy to do as it requires a list of disk names.
For AWS it needs a tuple (diskname,mountpoint) which would get nasty on the CLI or I take the list of disk names and loop through the whole alphabet for device mount points (/dev/sd[a-z]) depending on how many disks the user wants to mount.

wdyt?

giovannt0: Ah, yes, I had the same issue when writing a recipe for AWS in dftimewolf. 
I ended up doing what you describe (with a limited alphabet though, to match AWS requirements), see https://github.com/log2timeline/dftimewolf/blob/10573673319032bf08311938ae834ee329ec576d/dftimewolf/lib/collectors/aws.py#L233 (still a WIP with things to address, but adding for context).

hacktobeer: okay, cool. probably not going to something with a pre-populated list, popping and checking. if Amazon supports f-p it means we can connect 11 volumes at a maximum. easier to check on that

hacktobeer: kept the fix as simple as possible in the cli code.

giovannt0: This is great , thanks!

examples/aws_cli.py
```diff
@@ -107,3 +107,36 @@ def QueryLogs(args: 'argparse.Namespace') -> None:
...
+  Args:
+    args (argparse.Namespace): Arguments from ArgumentParser.
+  """
+  if len(args.attach_volumes) > 11:
+    print('error: --attach_volumes must be < 11')
+    return
+
+  attach_volumes = []
+  if args.attach_volumes:
+    volumes = args.attach_volumes.split(',')
```
tomchop, line 124: Split `args.attach_volumes` into elements of a list and do the check on that. 

```
volumes = args.attach_volumes.split(',')
# do extra cleanup steps such as avoiding ['']
if volumes:
  # blah
```

examples/aws_cli.py
```diff
@@ -107,3 +107,36 @@ def QueryLogs(args: 'argparse.Namespace') -> None:
...
+      device_letter = device_letter + 1
+
+  vm = forensics.StartAnalysisVm(vm_name=args.instance_name,
+                                 default_availability_zone=args.zone,
+                                 boot_volume_size=int(args.disk_size),
+                                 cpu_cores=int(args.cpu_cores),
+                                 ssh_key_name=args.ssh_key_name,
+                                 attach_volumes=attach_volumes)
+
+  print('Analysis VM started.')
```
tomchop, line 139: Maybe print a message before starting the VM as well?

examples/gcp_cli.py
```diff
@@ -97,3 +97,24 @@ def QueryLogs(args: 'argparse.Namespace') -> None:
...
+
+def StartAnalysisVm(args: 'argparse.Namespace') -> None:
+  """Start forensic analysis VM.
+
+  Args:
+    args (argparse.Namespace): Arguments from ArgumentParser.
+  """
+  attach_disks = []
+  if args.attach_disks:
+    attach_disks = args.attach_disks.split(',')
```
tomchop, line 109: Same, make sure the variables here are as clean as possible before passing them on to `forensics.StartAnalysisVM`

hacktobeer:
@tomchop PTAL.

Side note: Currently the input parameter are called attach_disks and attach_volumes for gcp/aws. This matches the terminology used by those cloud providers. Should be rename the aws parameter to attach_disks? I find that more consistent if we want to offer a uniform CLI tool which works as similar as possible across cloud provider.

giovannt0:
@hacktobeer If we rename this we should also rename `—volume_id` to stay consistent. So long as it’s clear in the docstring what these parameters reference I think this is a good idea to make the CLI tool more uniform.

hacktobeer:
agree, let me open a separate issue for this to keep this PR clean.

tomchop:
> @tomchop PTAL.
> 
> Side note: Currently the input parameter are called attach_disks and attach_volumes for gcp/aws. This matches the terminology used by those cloud providers. Should be rename the aws parameter to attach_disks? I find that more consistent if we want to offer a uniform CLI tool which works as similar as possible across cloud provider.

Yeah, I agree that it will be more consistent, but then one could argue we'd have to rename all the terms (i.e. passing `Volumes` as a `attach_disks` parameters might feel weird). I also wonder if down the line we'll end up spending too much time trying too hard to accommodate a common taxonomy of objects across cloud providers. I think it's fine to leave as-is for now, and let's further the discussion in a separate issue.

examples/gcp_cli.py
```diff
@@ -107,7 +107,15 @@ def StartAnalysisVm(args: 'argparse.Namespace') -> None:
...
   if args.attach_disks:
     attach_disks = args.attach_disks.split(',')
-
+    # Check if attach_disks parameter exists and if there
+    # are any empty entries.
+    if attach_disks and all(elements for elements in attach_disks):
+      pass
+    else:
+      print('error: parameter --attach_disks: {0:s}'.format(args.attach_disks))
+      return
```
tomchop, line 116: To make the code slightly simpler, you can use a guard statement:

```python
if not (attach_disks and all(elements for elements in attach_disks)):
  print(...)
  return

# Guard was passed proceed as normal
print('Starting analysis VM')
```

This is especially valuable here since you're just using `if / pass / else`, but also applies to `aws_cli.py`

examples/aws_cli.py
```diff
@@ -119,13 +119,20 @@ def StartAnalysisVm(args: 'argparse.Namespace') -> None:
...
-      device_letter = device_letter + 1
-
+      device_letter = ord('f')
+      for volume in volumes:
+        attach = (volume, '/dev/sd'+chr(device_letter))
+        attach_volumes.append(attach)
+        device_letter = device_letter + 1
+    else:
+      print('error: parameter --attach_volumes: {0:s}'.format(args.attach_volumes))
+      return
```
tomchop, line 136: See my comment about guard statements below.

PR #169, opened 2020-06-25
Import Disks From GCS ++

sa3eed3ed:
Adding Cloud Storage Functionalities
Adding Import function for Images stored in GCS
Adding Create disk from Image
Adding import disk from GCS functionality to gcp/forensics.py
Updating cli with import disk from GCS functionality
Adding unit tests fro Cloud Build and Cloud Storage
Fixing Doc strings


libcloudforensics/providers/gcp/internal/storage.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+
+    bucket, object_path = self.SplitGcsPath(gcs_path)
+    gcs_api_client = self.GcsApi().objects()
+    request = gcs_api_client.get(
+        bucket=bucket, object=object_path, userProject=user_project)
+    response = request.execute()  # type: Dict[str, Any]
+    return response
+
+  def GetMD5Object(
```
Fryyyyy, line 93: Instead of two functions that get the same thing and just return a different element of it, can we have a generic GetObjectMetadata or something and then the callee can just access whatever attribute they care about? This will make my life a lot easier on the internal side...

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -759,31 +920,30 @@ def GetOperation(self) -> Dict[str, Any]:
...
     """Export compute image to Google Cloud storage.
 
     Exported image is compressed and stored in .tar.gz format.
 
     Args:
       gcs_output_folder (str): Folder path of the exported image.
-      output_name (str): Optional. Name of the output file, must end with
-          .tar.gz, if not exist, the [image_name].tar.gz will be used.
+      output_name (str): Optional. Name of the output file. Name will be
+          appeneded with .tar.gz. Default is [image_name].tar.gz.
```
giovannt0, line 936: Typo, `appended`

libcloudforensics/providers/gcp/internal/compute_base_resource.py
```diff
@@ -76,7 +84,7 @@ def GetValue(self, key: str) -> str:
     """
 
     # pylint: disable=no-member
-    self._data = self.GetOperation()  # type: ignore
+    self._data = self.GetOperation()   # type: ignore
```
giovannt0, line 87: Do we still need the pylint directive `no-member` and mypy `ignore` with the newly defined abstract method line 68?

libcloudforensics/providers/gcp/internal/project.py
```diff
@@ -60,10 +55,11 @@ def __init__(self,
     self._function = None
     self._build = None
     self._log = None
+    self._storage = None
     self._monitoring = None
 
   @property
-  def compute(self) -> compute_module.GoogleCloudCompute:
+  def compute(self) -> 'compute_module.GoogleCloudCompute':
```
giovannt0, line 62: Why add the quotes `''` here for the return types? It seems that mypy was happy without them before (afaik you only need the quotes when the type has not be defined yet (see https://www.python.org/dev/peps/pep-0484/#forward-references)) 

Did something change in the way things are imported?

Same for following methods

libcloudforensics/providers/gcp/internal/project.py
```diff
@@ -119,7 +115,21 @@ def log(self) -> log_module.GoogleCloudLog:
     return self._log  # type: ignore
 
   @property
-  def monitoring(self) -> monitoring_module.GoogleCloudMonitoring:
+  def storage(self) -> 'storage_module.GoogleCloudStorage':
+    """Get a GoogleCloudLog object for the project.
```
giovannt0, line 119: Docstring needs to be updated from Log to Storage

tests/providers/gcp/gcp_test.py
```diff
@@ -634,8 +695,75 @@ def testExecuteQuery(self, mock_gcl_api):
...
 
+class GoogleCloudStorageTest(unittest.TestCase):
+  """Test Google Cloud Storage class."""
+  # pylint: disable=line-too-long
+
+  @typing.no_type_check
+  def testSplitGcsPath(self):
+    """Tests that GCS path split is correctly done."""
+    bucket, object_uri = FAKE_GCS.SplitGcsPath(FAKE_GCS_PATH)
+    self.assertEqual(object_uri, FAKE_GCS_URI)
```
giovannt0, line 706: In this test and the followings, please format equal assertions as `assertEqual(expected, actual)` and use the raw value if known/if possible:

```
self.assertEqual('fake-folder/fake-object', object_uri)
```

This makes it easier to read and know what to expect when reading through the test file :)

tests/providers/gcp/gcp_test.py
```diff
@@ -634,8 +695,75 @@ def testExecuteQuery(self, mock_gcl_api):
...
+  @typing.no_type_check
+  @mock.patch('libcloudforensics.providers.gcp.internal.build.GoogleCloudBuild.GcbApi')
+  def testBlockOperation(self, mock_gcb_api):
+    """Test that Cloud Builds are correctly blocked until done."""
+    build_operation_object = mock_gcb_api.return_value.operations.return_value.get
+    build_operation_object.return_value.execute.return_value = MOCK_GCB_BUILDS_SUCCESS
+    block_build_success = FAKE_GCB.BlockOperation(MOCK_GCB_BUILDS_CREATE)
+    self.assertEqual(block_build_success, MOCK_GCB_BUILDS_SUCCESS)
+    build_operation_object.return_value.execute.return_value = MOCK_GCB_BUILDS_FAIL
+    with self.assertRaises(RuntimeError):
```
giovannt0, line 761: To stay consistent with other tests, wdyt about changing this to:

```
self.assertRaises(RuntimeError, FAKE_GCB.BlockOperation, MOCK_GCB_BUILDS_CREATE)
```

I don't know if this makes things better or worst to be honest, we could also change the other assertions to the `with` format in a separate PR. Thoughts @tomchop @hacktobeer ?

libcloudforensics/providers/gcp/internal/storage.py
tomchop: I agree, `GetMD5Object` and `GetSizeObject` are just wrapping `GetOperationObject` with very limited functionality (mostly accessing keys in the returned dict). Can we rename `GetOperationObject` to something that sounds more user-friendly (e.g. like @Fryyyyy suggested `GetObjectMetadata`) and have the callers deal with whatever content they're interested in?

libcloudforensics/providers/gcp/internal/storage.py
```diff
@@ -0,0 +1,132 @@
...
+      gcs_path (str): File path to a resource in GCS.
+          Ex: gs://bucket/folder/obj
+
+    Returns:
+      Tuple[str, str]: Bucket name. Object URI.
+    """
+
+    _, _, full_path = gcs_path.partition('//')
+    bucket, _, object_uri = full_path.partition('/')
+    return bucket, object_uri
```
tomchop, line 69: `self` is never used here, so this should probably be a static method outside the class.

tests/providers/gcp/gcp_test.py
```diff
@@ -634,8 +695,75 @@ def testExecuteQuery(self, mock_gcl_api):
...
+  @typing.no_type_check
+  @mock.patch('libcloudforensics.providers.gcp.internal.storage.GoogleCloudStorage.GcsApi')
+  def testGetMD5Object(self, mock_gcs_api):
+    """Test GCS object MD5 hash."""
+    api_get_object = mock_gcs_api.return_value.objects.return_value.get
+    api_get_object.return_value.execute.return_value = MOCK_GCS_OPERATION_OBJECT_GET
+    md5_base64 = FAKE_GCS.GetMD5Object('gs://Fake_Path')
+    self.assertEqual(md5_base64, MOCK_GCS_OPERATION_OBJECT_GET['md5Hash'])
+    md5_hex = FAKE_GCS.GetMD5Object('gs://Fake_Path', in_hex=True)
+    self.assertEqual(md5_hex, base64.b64decode(MOCK_GCS_OPERATION_OBJECT_GET['md5Hash']).hex())
```
tomchop, line 727: When testing, please always be as explicit as possible in your comparisons and use literals whenever possible, e.g.

```python
self.assertEqual(md5_hex, 'abcdef....')
```

This makes the test much more clear for anyone wondering what `md5_hex` is expected to contain, and also prevents bugs in your test (in this case, it seems that `MOCK_GCS_OPERATION_OBJECT_GET['md5Hash']`is not a base64 string (`fakehash`), so I wonder how `b64decode` will work on that)

(Bringing this to @binglot's attention as we talked about testing before)

tests/providers/gcp/gcp_test.py
```diff
@@ -634,8 +695,75 @@ def testExecuteQuery(self, mock_gcl_api):
...
+
+  @typing.no_type_check
+  @mock.patch('libcloudforensics.providers.gcp.internal.storage.GoogleCloudStorage.GcsApi')
+  def testGetSizeObject(self, mock_gcs_api):
+    """Test GCS object size."""
+    api_get_object = mock_gcs_api.return_value.objects.return_value.get
+    api_get_object.return_value.execute.return_value = MOCK_GCS_OPERATION_OBJECT_GET
+    object_size = FAKE_GCS.GetSizeObject('gs://Fake_Path')
+    self.assertEqual(object_size, MOCK_GCS_OPERATION_OBJECT_GET['size'])
```
tomchop, line 736: Same comment as above, if we could use the actual integer value that would be better. If you want to be explicit as to why it should be that value, you can add a comment saying where to find it.

tests/providers/gcp/gcp_test.py
```diff
@@ -184,6 +192,59 @@
     }]
 }
 
+MOCK_GCS_OPERATION_OBJECT_GET = {
```
tomchop, line 195: Rename this to MOCK_OBJECT_METADATA if you decide to rename the function as mentioned above.

tests/providers/gcp/gcp_test.py
tomchop: +1, see comments below too

examples/gcp_cli.py
```diff
@@ -102,6 +102,24 @@ def QueryLogs(args: 'argparse.Namespace') -> None:
     print(json.dumps(line))
 
 
+def CreateDiskFromGCSImage(args: 'argparse.Namespace') -> None:
+  """Creates GCE persistent disk from image in GCS.
+
+  Args:
+      args (argparse.Namespace): Arguments from ArgumentParser.
+  """
```
tomchop, line 110: Please point to the docstring for `CreateDiskFromGCSImage` to get the details about how the image is created.

libcloudforensics/providers/gcp/forensics.py
```diff
@@ -139,3 +140,73 @@ def StartAnalysisVm(
...
+          'project_id': 'fake-project',
+          'disk_name': 'fake-imported-disk',
+          'zone': 'fake-zone',
+          'bytes_count': '1234'  # Content-Length of source image in bytes.
+          'md5Hash': 'Source Image MD5 hash string in hex'
+        }
+
+  Raises:
+    RuntimeError: If GCE Disk name is invalid.
+  """
```
tomchop, line 187:
```suggestion
  """Creates a GCE persistent disk from a image in GCS.

  The method supports raw disk images and most virtual disk
  file formats. Valid import formats are:
  [raw (dd), qcow2, qcow , vmdk, vdi, vhd, vhdx, qed, vpc].

  The created GCE disk might be larger than the original raw (dd)
  image stored in GCS to satisfy GCE capacity requirements:
  https://cloud.google.com/compute/docs/disks/#introduction
  However the bytes_count and the md5_hash values of the source
  image are returned with the newly created disk.
  The md5_hash can be used to verify the integrity of the
  created GCE disk, it must be compared with the hash of the
  created GCE disk from byte 0 to bytes_count. i.e:
  result['md5Hash'] = hash(created_gce_disk,
                            start_byte=0,
                            end_byte=result['bytes_count'])

  Args:
    project_id (str): Google Cloud Project ID.
    storage_image_path (str): Path to the source image in GCS.
    zone (str): Zone to create the new disk in.
    name (str): Optional. Name of the disk to create. Default
        is imported-disk-[TIMESTAMP('%Y%m%d%H%M%S')].

  Returns:
    Dict: A key value describing the imported GCE disk.
        Ex: {
          'project_id': 'fake-project',
          'disk_name': 'fake-imported-disk',
          'zone': 'fake-zone',
          'bytes_count': '1234'  # Content-Length of source image in bytes.
          'md5Hash': 'Source Image MD5 hash string in hex'
        }

  Raises:
    RuntimeError: If the GCE disk name is invalid.
  """
```

libcloudforensics/providers/gcp/forensics.py
```diff
@@ -139,3 +140,73 @@ def StartAnalysisVm(
...
+
+  if name:
+    if not common.REGEX_DISK_NAME.match(name):
+      raise RuntimeError(
```
tomchop, line 191: It would be better to raise a `ValueError` here

libcloudforensics/providers/gcp/internal/storage.py
```diff
@@ -0,0 +1,132 @@
...
+
+  def GetOperationObject(
+      self, gcs_path: str,
+      user_project: Optional[str] = None) -> Dict[str, Any]:
```
tomchop, line 73:
```suggestion
  def GetOperationObject(self,
                         gcs_path: str,
                         user_project: Optional[str] = None) -> Dict[str, Any]:
```

tests/providers/gcp/gcp_test.py
sa3eed3ed: as discussed let's use assertRaises as a context manager (+ with key word) it makes it more explicit and the others we change in issue/173.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -467,27 +463,41 @@ def __ListByLabel(
           previous_request=request, previous_response=response)
     return resource_dict
 
-  def CreateImageFromDisk(self,
-                          src_disk: 'GoogleComputeDisk',
-                          name: Optional[str] = None) -> 'GoogleComputeImage':
+  def CreateImageFromDisk(
+      self, src_disk: 'GoogleComputeDisk',
```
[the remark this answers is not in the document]
sa3eed3ed: this is all yapf stuff, I ran it by mistake and later I had to undo all the changes, but few things went overlooked, will undo these

libcloudforensics/providers/gcp/internal/storage.py
sa3eed3ed: yes, changed GetOperationObject to get Object metadata, removed GetObjectSize and accessed the size element directly. Left GetObjectMD5 function there since it handles the different encoding of the returned hash. 
PS: We have similar implementation in compute_base_resource.py which I had the same concerns about, it make sense to remove all these methods there and access the element needed directly from the getValue function.

libcloudforensics/providers/gcp/forensics.py
```diff
@@ -139,3 +140,73 @@ def StartAnalysisVm(
...
+    name = 'imported-disk-{0}'.format(timestamp)
+
+  result = {}
+  project = gcp_project.GoogleCloudProject(project_id)
+  image_object = project.compute.ImportImageFromStorage(storage_image_path)
+  disk_object = project.compute.CreateDiskFromImage(
+      image_object, zone=zone, name=name)
+  bytes_count = project.storage.GetSizeObject(storage_image_path)
+  md5_hash = project.storage.GetMD5Object(storage_image_path, in_hex=True)
+  result['project_id'] = disk_object.project_id
```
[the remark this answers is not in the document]
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -57,7 +55,8 @@ def __init__(self,
     super(GoogleCloudCompute, self).__init__(self.project_id)
 
   def Instances(self,
-                refresh: bool = True) -> Dict[str, 'GoogleComputeInstance']:
+                refresh: Optional[bool] = True
+               ) -> Dict[str, 'GoogleComputeInstance']:
```
giovannt0, line 59: Probably some more yapf stuff

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -73,7 +72,8 @@ def Instances(self,
     return self._instances  # type: ignore
 
   def Disks(self,
-            refresh: bool = True) -> Dict[str, 'GoogleComputeDisk']:
+            refresh: Optional[bool] = True
```
giovannt0, line 75: Probably some more yapf stuff

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -231,21 +231,20 @@ def CreateDiskFromSnapshot(
...
-      image_family: str = 'ubuntu-1804-lts',
-      # pylint: disable=line-too-long
-      packages: Optional[List[str]] = None) -> Tuple['GoogleComputeInstance', bool]:
-      # pylint: enable=line-too-long
+    return GoogleComputeDisk(
+        project_id=self.project_id,
+        zone=self.default_zone,  # type: ignore
+        name=disk_name)
+
+  def GetOrCreateAnalysisVm(self,
```
giovannt0, line 239: Probably some more yapf stuff

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -347,10 +344,10 @@ def GetOrCreateAnalysisVm(
     created = True
     return instance, created
 
-  def ListInstanceByLabels(
-      self,
-      labels_filter: Dict[str, str],
-      filter_union: bool = True) -> Dict[str, 'GoogleComputeInstance']:
+  def ListInstanceByLabels(self,
```
giovannt0, line 347: Probably some more yapf stuff

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -371,12 +368,12 @@ def ListInstanceByLabels(
...
     instance_service_object = self.GceApi().instances()
     return self.__ListByLabel(  # type: ignore
-        labels_filter, instance_service_object, filter_union)
+        labels_filter, instance_service_object, filter_union)  # type: ignore
 
-  def ListDiskByLabels(
-      self,
-      labels_filter: Dict[str, str],
-      filter_union: bool = True) -> Dict[str, 'GoogleComputeDisk']:
+  def ListDiskByLabels(self,
```
giovannt0, line 373: Probably some more yapf stuff

sa3eed3ed: for consistency with other functions

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: for consistency with other functions

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: for consistency with other fucntions

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: to avoid line too long

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: so these ones, to avoid adding the long-line exception thing, I actually moved it to a new line otherwise it is line too long

examples/gcp_cli.py
sa3eed3ed: Done

tests/providers/gcp/gcp_test.py
sa3eed3ed: Done

tests/providers/gcp/gcp_test.py
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/compute.py
giovannt0: Ah, I see, nevermind then :)

tests/providers/gcp/gcp_test.py
sa3eed3ed: Done. Actually b64decode works, it doesn't error and that is the relevant things for this test.

libcloudforensics/providers/gcp/internal/storage.py
sa3eed3ed: Done

tests/providers/gcp/gcp_test.py
```diff
@@ -634,8 +692,67 @@ def testExecuteQuery(self, mock_gcl_api):
...
+  @typing.no_type_check
+  @mock.patch('libcloudforensics.providers.gcp.internal.storage.GoogleCloudStorage.GcsApi')
+  def testGetMD5Object(self, mock_gcs_api):
+    """Test GCS object MD5 hash."""
+    api_get_object = mock_gcs_api.return_value.objects.return_value.get
+    api_get_object.return_value.execute.return_value = MOCK_GCS_OBJECT_METADATA
+    md5_base64 = FAKE_GCS.GetMD5Object('gs://Fake_Path')
+    self.assertEqual('fakehash', md5_base64)
+    md5_hex = FAKE_GCS.GetMD5Object('gs://Fake_Path', in_hex=True)
+    self.assertEqual(base64.b64decode('fakehash').hex(), md5_hex)
```
giovannt0, line 725: I'd suggest `self.assertEqual('7da91e85ab21', md5_hex)` which is the decoded value expected

tests/providers/gcp/gcp_test.py
```diff
@@ -634,8 +692,67 @@ def testExecuteQuery(self, mock_gcl_api):
...
+  # pylint: disable=line-too-long
+
+  @typing.no_type_check
+  @mock.patch('libcloudforensics.providers.gcp.internal.build.GoogleCloudBuild.GcbApi')
+  def testCreateBuild(self, mock_gcb_api):
+    """Test that Cloud Builds are correctly created."""
+    build_create_object = mock_gcb_api.return_value.projects.return_value.builds.return_value.create
+    build_create_object.return_value.execute.return_value = MOCK_GCB_BUILDS_CREATE
+    build_response = FAKE_GCB.CreateBuild({'Fake-Build_body': None})
+    self.assertEqual(build_response, MOCK_GCB_BUILDS_CREATE)
```
giovannt0, line 739: `self.assertEqual(MOCK_GCB_BUILDS_CREATE, build_response)` to conform to `(expected, actual)`

libcloudforensics/providers/gcp/internal/storage.py
sa3eed3ed: let me know if this make sense

tests/providers/gcp/gcp_test.py
sa3eed3ed: done

tests/providers/gcp/gcp_test.py
sa3eed3ed: done

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -497,6 +507,152 @@ def CreateImageFromDisk(self,
...
+
+    Returns:
+      GoogleComputeImage: A Google Compute Image object.
+
+    Raises:
+      RuntimeError: If bootable is True and os_name not specified or
+          if imported image name is invalid.
+    """
+
+    supported_os = \
```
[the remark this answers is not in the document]
sa3eed3ed: not directly it is all part of Daisy troggered by cloud build: https://googlecloudplatform.github.io/compute-image-tools/image-import.html

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -497,6 +507,152 @@ def CreateImageFromDisk(self,
...
+        raise RuntimeError(
+            'Imported image name {0:s} does not comply with {1:s}'.format(
+                image_name, common.REGEX_DISK_NAME.pattern))
+      truncate_at = 63
+      image_name = image_name[:truncate_at]
+    else:
+      timestamp = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+      image_name = 'imported-image-{0}'.format(timestamp)
+    args_list = [
+        '-image_name={image_name}'.format(image_name=image_name),
```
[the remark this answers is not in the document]
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -497,6 +507,152 @@ def CreateImageFromDisk(self,
...
+    if image_name:
+      if not common.REGEX_DISK_NAME.match(image_name):
+        raise RuntimeError(
+            'Imported image name {0:s} does not comply with {1:s}'.format(
+                image_name, common.REGEX_DISK_NAME.pattern))
+      truncate_at = 63
+      image_name = image_name[:truncate_at]
+    else:
+      timestamp = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+      image_name = 'imported-image-{0}'.format(timestamp)
```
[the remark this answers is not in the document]
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -497,6 +507,152 @@ def CreateImageFromDisk(self,
...
+    args_list = [
+        '-image_name={image_name}'.format(image_name=image_name),
+        '-source_file={source_file}'.format(source_file=storage_image_path),
+        '-timeout=86400s',
+        '-client_id=api',
+        '{img_type}'.format(img_type=img_type)
```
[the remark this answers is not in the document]
sa3eed3ed: done

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -497,6 +507,152 @@ def CreateImageFromDisk(self,
...
+    ]
+    if guest_env:
+      args_list.append('{guest_env}'.format(guest_env=guest_env))
```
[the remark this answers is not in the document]
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: good catch!

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -497,6 +507,152 @@ def CreateImageFromDisk(self,
...
+    if not bootable:
+      img_type = '-data_disk'
+    elif not os_name:
+      raise RuntimeError(
+          'For bootable images, operating system name'
+          ' (os_name) must be specified.')
+    elif os_name not in supported_os:
+      common.LOGGER.warning(
+          ('Operating system of the imported image is not within the '
+           'supported list:\n{0:s}\nFor the uptodate list please refer '
```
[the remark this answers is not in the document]
sa3eed3ed: done

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -467,27 +463,41 @@ def __ListByLabel(
...
     Returns:
       GoogleComputeImage: A Google Compute Image object.
+
+    Raises:
+      RuntimeError: If GCE Image name is invalid.
     """
-    if not name:
-      name = '{0:s}_image'.format(src_disk.name)
+
+    truncate_at = 63
```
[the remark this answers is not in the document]
sa3eed3ed: done, added StampAndTruncateName to common and referenced it in forensics.py and compute.py

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -497,6 +507,152 @@ def CreateImageFromDisk(self,
...
+          ('Operating system of the imported image is not within the '
+           'supported list:\n{0:s}\nFor the uptodate list please refer '
+           'to:\n{1:s}').format(
+               ', '.join(supported_os),
+               'https://cloud.google.com/sdk/gcloud/reference/compute/images/import#--os'))  # pylint: disable=line-too-long
+    else:
+      img_type = '-os={0}'.format(os_name)
+      if guest_environment:
+        guest_env = ''
+    if image_name:
```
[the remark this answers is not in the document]
sa3eed3ed: done

libcloudforensics/providers/gcp/internal/storage.py
```diff
@@ -0,0 +1,114 @@
...
+    """Get API operation object metadata for Google Cloud Storage object.
+
+    Args:
+      gcs_path (str): File path to a resource in GCS.
+          Ex: gs://bucket/folder/obj
+      user_project (str): The project ID to be billed for this request.
+          Required for Requester Pays buckets.
+
+    Returns:
+      Dict: An API operation object for a Google Cloud Storage object.
```
Fryyyyy, line 83: Could be useful to link to https://cloud.google.com/storage/docs/json_api/v1/objects#resource since it has the key/value information for what will actually be returned - "API operation object" is pretty vague

tests/providers/gcp/gcp_test.py
```diff
@@ -722,7 +721,8 @@ def testGetMD5Object(self, mock_gcs_api):
     md5_base64 = FAKE_GCS.GetMD5Object('gs://Fake_Path')
     self.assertEqual('fakehash', md5_base64)
     md5_hex = FAKE_GCS.GetMD5Object('gs://Fake_Path', in_hex=True)
-    self.assertEqual(base64.b64decode('fakehash').hex(), md5_hex)
+    # '7da91e85ab21' = base64.b64decode('fakehash').hex()
+    self.assertEqual('7da91e85ab21', md5_hex)
```
tomchop, line 725: I guess what I meant here is that the string `fakehash` is not a great choice to use as a test value because you end up comparing `7da91e85ab21` to an md5 value, and this may be confusing to someone trying to figure out what md5_hex contains. Better use a value that makes semantic sense (a b64 encoding of an actual md5 hash)

libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -85,6 +86,26 @@ def GenerateDiskName(snapshot: 'compute.GoogleComputeSnapshot',
   return disk_name
 
 
+def StampAndTruncateName(prefix: str,
```
tomchop, line 89: I'd use a more generic name e.g. `GenerateUniqueInstanceName` to separate the naming from the logic of the function. The docstring is more than enough to figure out what's going on behind the scenes.

libcloudforensics/providers/gcp/internal/compute_base_resource.py
```diff
@@ -65,13 +65,13 @@ def FormatLogMessage(self, message: str) -> str:
 
     return 'project:{0} {1}'.format(self.project_id, message)
 
-  def GetOperation(self) -> Dict[str, Any]:
+  def GetOperation(self) -> 'Dict[str, Any]':
```
tomchop, line 68: Why are we using quotes here? The other `Dict` type checks (see L39) don't have them

tests/providers/gcp/gcp_test.py
```diff
@@ -702,18 +699,19 @@ class GoogleCloudStorageTest(unittest.TestCase):
   @typing.no_type_check
   def testSplitGcsPath(self):
     """Tests that GCS path split is correctly done."""
-    bucket, object_uri = FAKE_GCS.SplitGcsPath(FAKE_GCS_PATH)
-    self.assertEqual(object_uri, FAKE_GCS_URI)
-    self.assertEqual(bucket, FAKE_GCS_BUCKET)
+    bucket, object_uri = FAKE_GCS.SplitGcsPath('gs://fake-bucket/fake-folder/fake-object')
```
tomchop: Probably line-too-long?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -497,6 +506,146 @@ def CreateImageFromDisk(self,
...
+
+    Returns:
+      GoogleComputeImage: A Google Compute Image object.
+
+    Raises:
+      ValueError: If bootable is True and os_name not specified or
+          if imported image name is invalid.
+    """
+
+    supported_os = \
```
tomchop, line 587: You don't need the backslash here:

```python
supported_os = [
  blah1, blah2, etc.
]
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -432,7 +428,8 @@ def __ListByLabel(
       raise RuntimeError(error_msg)
 
     # pylint: disable=line-too-long
-    resource_dict = {}  # type: Dict[str, Union[GoogleComputeInstance, GoogleComputeDisk]]
+    resource_dict = {
+    }  # type: Dict[str, Union[GoogleComputeInstance, GoogleComputeDisk]]
```
tomchop, line 432: If line-too-long no need to split the dict :D

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -475,19 +472,31 @@ def CreateImageFromDisk(self,
...
       src_disk (GoogleComputeDisk): Source disk for the image.
       name (str): Optional. Name of the image to create. Default
-          is [src_disk]_image.
+          is [src_disk.name]-[TIMESTAMP('%Y%m%d%H%M%S')].
 
     Returns:
       GoogleComputeImage: A Google Compute Image object.
+
+    Raises:
+      ValueError: If GCE Image name is invalid.
```
tomchop, line 481:
```suggestion
      ValueError: If the GCE Image name is invalid.
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -570,9 +719,16 @@ def __SshConnection(self) -> None:
...
+        'gcloud',
+        'compute',
+        '--project',
+        self.project_id,
+        'ssh',
+        '--zone',
+        self.zone,
+        self.name
+    ],
+                          stderr=devnull)
```
tomchop, line 731: plz fix indentation here. Might be a good idea to build the command list in one variable and have the full `check_call` call in one line

libcloudforensics/providers/gcp/internal/storage.py
```diff
@@ -90,11 +90,10 @@ def GetOperationObject(
     response = request.execute()  # type: Dict[str, Any]
     return response
 
-  def GetMD5Object(
-      self,
-      gcs_path: str,
-      user_project: Optional[str] = None,
-      in_hex: Optional[bool] = False) -> str:
+  def GetMD5Object(self,
```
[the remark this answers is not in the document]
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/storage.py
sa3eed3ed: Done

tests/providers/gcp/gcp_test.py
sa3eed3ed: done

libcloudforensics/providers/gcp/internal/common.py
sa3eed3ed: Done

tests/providers/gcp/gcp_test.py
sa3eed3ed: I disabled line-too-long after starting the GoogleCloudStorageTest class, but this is already done everywhere in the gcp_test module

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: that is definitely a remnant of  my "out of mistake" yapf check that I undid manually

sa3eed3ed: Fixed it

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/compute_base_resource.py
sa3eed3ed: removed

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: done

libcloudforensics/providers/gcp/internal/build.py
```diff
@@ -89,17 +90,34 @@ def BlockOperation(self, response: Dict[str, Any]) -> Dict[str, Any]:
...
+          logging.info(
+              'build.BlockOperation: Get request to cloudbuild.googleapis.com '
+              'failed.\nTry {0:d} of {1:d}. Error: {2!s} '.format(
+                  block_retry, BLOCK_RETRY_MAX, error))
+        if get_success:
+          break
+        if block_retry == BLOCK_RETRY_MAX - 1:
+          raise RuntimeError(
+              'Faliure blocking Cloud Build operation: {0:s}'.format(
+                  response['name']))
```
tomchop, line 114: This function is pretty complex and has two levels of retries, one where we catch HTTP errors, limited to a MAX number and the other where we catch other types of errors, not capped (so will retry until a response is obtained.

I think it would make sense in this case to have the retry loops in different functions (just add a new one to do the capped http exception catching, something like _RetryExecuteRequest()). No need to declare the BLOCK_RETRY_MAX as a module-level variable since it's only used by this function. Also consider using the `for / else` construct which might be of use here.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -397,13 +394,12 @@ def ListDiskByLabels(
...
     return self.__ListByLabel(  # type: ignore
-        labels_filter, disk_service_object, filter_union)
+        labels_filter, disk_service_object, filter_union)  # type: ignore
 
-  def __ListByLabel(
-      self,
-      labels_filter: Dict[str, str],
-      service_object: 'googleapiclient.discovery.Resource',
-      filter_union: bool) -> Dict[str, Union['GoogleComputeInstance', 'GoogleComputeDisk']]:  # pylint: disable=line-too-long
+  def __ListByLabel(self,
```
tomchop, line 399: Use single leading underscores if you want to declare private functions (also in `__SshConnection`)

libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -35,6 +35,7 @@
 LOGGER = logging.getLogger()
 RETRY_MAX = 10
 REGEX_DISK_NAME = re.compile('^(?=.{1,63}$)[a-z]([-a-z0-9]*[a-z0-9])?$')
+COMPUTE_NAME_LIMIT = 63
```
tomchop, line 38: shouldn't this go in `compute.py`?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -569,10 +718,15 @@ def __SshConnection(self) -> None:
...
-    ], stderr=devnull)
+    cmd_list = ['gcloud',
+                'compute',
+                '--project',
+                self.project_id,
+                'ssh',
+                '--zone',
+                self.zone,
+                self.name]
+    subprocess.check_call(cmd_list, stderr=devnull)
```
tomchop, line 729: Does this really need its own function? Can't you just call it from `Ssh`?

libcloudforensics/providers/gcp/internal/common.py
sa3eed3ed: this is for the instance name generation method

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: this doesn't belong to this pull request at all it was there from before. I just moved the cmds to a list as per your previous request, we can maybe take this to a another issue? I believe in general SSHing is not the best solution for on instance execution and startup script is more stable way to do that

tomchop: Yes agreed.

libcloudforensics/providers/gcp/internal/build.py
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: Done. Neither of these methods belong to this PR, if they were touched it is the yapf stuff, I changed them in all places, they can't be referenced from other places so should be fine

tomchop:
CLA is good, all authors are Googlers.

PR #174, opened 2020-06-26
Forensic disk copy capabilities in Azure

giovannt0:
Bringing Azure to the party with this PR! It follows the other providers structure and offers at the moment only disk copying functionality. It should be noted that this is an initial PR with base functionality. Creating copies in different regions or Azure account / subscriptions is not yet supported.

- CLI: #181 
- Unit tests: #179 
- E2E: #185 

Authentication: As explained in https://docs.microsoft.com/en-us/azure/developer/python/azure-sdk-authenticate?tabs=cmd, they have some trouble making `DefaultAuthentication` work with the new management libraries. Therefore at the moment, authentication is done through environment variables that users have to set up according to the documentation.

Happy Friday :)


libcloudforensics/providers/azure/forensics.py
```diff
@@ -0,0 +1,69 @@
...
+
+if TYPE_CHECKING:
+  from libcloudforensics.providers.azure.internal import compute
+
+
+def CreateDiskCopy(
+    subscription_id: str,
+    instance_name: str,
+    disk_name: Optional[str] = None,
+    disk_type: Optional[str] = 'Standard_LRS') -> 'compute.AZDisk':
```
hacktobeer, line 30: remove quotes from return type...

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,221 @@
...
+      subscription_id (str): The Azure subscription ID to use.
+    """
+    self.subscription_id = subscription_id
+    self.credentials = common.GetCredentials()
+    self.compute_client = ComputeManagementClient(
+        self.credentials, self.subscription_id)
+
+  def ListInstances(
+      self,
+      # pylint: disable=line-too-long
```
hacktobeer, line 51: this makes it unreadable for me, had to look 4 times. just 1 time at the end of the long line makes it (for me at least) readable again.

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,221 @@
...
+        instances[vm_name] = compute.AZVirtualMachine(self,
+                                                      instance.id,
+                                                      vm_name,
+                                                      instance.location,
+                                                      zones=instance.zones)
+    return instances
+
+  def ListDisks(
+      self,
+      # pylint: disable=line-too-long
```
hacktobeer, line 86: :(

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,221 @@
...
+
+    Returns:
+      Dict[str, AZDisk]: Dictionary mapping disk names (str) to their
+          respective AZDisk object.
+    """
+    disks = {}  # type: Dict[str, compute.AZDisk]
+    az_disk_client = self.compute_client.disks
+    if not resource_group_name:
+      responses = common.ExecuteRequest(az_disk_client, 'list')
+    else:
```
hacktobeer, line 104: **kwargs??

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,221 @@
...
+      responses = common.ExecuteRequest(
+          az_disk_client,
+          'list_by_resource_group',
+          {'resource_group_name': resource_group_name})  # type: ignore
+    for response in responses:
+      for disk in response:
+        disk_name = disk.name
+        disks[disk_name] = compute.AZDisk(self,
```
hacktobeer, line 112: disks[disk.name] instead of disks[disk_name]?

libcloudforensics/providers/azure/forensics.py
giovannt0:
```
************* Module libcloudforensics.providers.azure.forensics
libcloudforensics/providers/azure/forensics.py:30:50: E0601: Using variable 'compute' before assignment (used-before-assignment)
```

Happens without the quotes

libcloudforensics/providers/azure/internal/account.py
giovannt0: Yeah, not very nice within functions. Maybe we can do it this way for statements / urls but keep it on the line for functions?

libcloudforensics/providers/azure/internal/account.py
giovannt0: The functions called are different (not only the parameters), not sure how I can use **kwargs there

libcloudforensics/providers/azure/internal/account.py
giovannt0: Changed

libcloudforensics/providers/azure/forensics.py
```diff
@@ -0,0 +1,74 @@
...
+
+if TYPE_CHECKING:
+  from libcloudforensics.providers.azure.internal import compute
+
+
+def CreateDiskCopy(
+    subscription_id: str,
+    instance_name: Optional[str] = None,
+    disk_name: Optional[str] = None,
+    disk_type: str = 'Standard_LRS') -> 'compute.AZDisk':
```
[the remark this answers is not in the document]
giovannt0: We're not actually importing the `compute` package since we only use `AZDisk` for type hinting. We therefore use forward reference (see [PEP 484](https://www.python.org/dev/peps/pep-0484/#forward-references)) to declare the type. Removing the quotes will break `mypy`.

giovannt0:
Thanks @binglot, addressed

libcloudforensics/providers/azure/forensics.py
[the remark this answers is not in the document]
binglot: Thanks for the explanation Theo!

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,217 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Represents an Azure account."""
+
+from time import sleep
+from typing import Optional, Dict
+
+# pylint: disable=import-error
```
tomchop, line 20: Could you add a comment about why we're ignoring the `import-error` here?

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,217 @@
...
+      subscription_id (str): The Azure subscription ID to use.
+    """
+    self.subscription_id = subscription_id
+    self.credentials = common.GetCredentials()
+    self.compute_client = ComputeManagementClient(
+        self.credentials, self.subscription_id)
+
+  def ListInstances(
+      self,
+      resource_group_name: Optional[str] = None) -> Dict[str, compute.AZVirtualMachine]:  # pylint: disable=line-too-long
```
tomchop, line 51: What about splitting this in two lines?

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,217 @@
...
+    """List instances in an Azure subscription / resource group.
+
+    Args:
+      resource_group_name (str): Optional. The resource group name to list
+          instances from. If none specified, then all instances in the AZ
```
tomchop, line 56: It's best to use the full name of the service when referred to in docstrings.

```suggestion
          instances from. If none specified, then all instances in the Azure
```

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,217 @@
...
+# pylint: disable=import-error
+from azure.mgmt.compute import ComputeManagementClient
+from azure.mgmt.compute.v2016_04_30_preview.models import DiskCreateOption
+from msrestazure.azure_exceptions import CloudError
+# pylint: enable=import-error
+
+from libcloudforensics.providers.azure.internal import compute, common
+
+
+class AZAccount:
```
tomchop, line 29: Any reason we're calling this `AZ*` throughout the codebase? If it's not an official product name or a widely adopted acronym, I'd prefer we stick to the full `Azure`

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    try:
+      request = self.compute_client.disks.create_or_update(
+          snapshot.resource_group_name,
+          disk_name,
+          creation_data,
+          sku=disk_type)
+      while not request.done():
+        sleep(5)  # Wait 5 seconds before checking disk status again
```
tomchop, line 207: Should we define a maximum wait time? This will loop forever otherwise

libcloudforensics/providers/azure/internal/account.py
tomchop: You seem to be calling `common.ExecuteRequest` both times, but I agree it's clearer to have two separate calls since the parameters are very different.

libcloudforensics/providers/azure/internal/common.py
```diff
@@ -0,0 +1,132 @@
...
+
+if TYPE_CHECKING:
+  # TYPE_CHECKING is always False at runtime, therefore it is safe to ignore
+  # the following cyclic import, as it it only used for type hints
+  from libcloudforensics.providers.azure.internal import compute  # pylint: disable=cyclic-import
+
+# pylint: disable=line-too-long
+# See https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules
+# pylint: enable=line-too-long
+REGEX_DISK_NAME = re.compile('^[\\w]{1,80}$')
```
tomchop, line 33:
```suggestion
REGEX_DISK_NAME = re.compile('^\w{1,80}$')
```

libcloudforensics/providers/azure/internal/common.py
```diff
@@ -0,0 +1,132 @@
...
+REGEX_SNAPSHOT_NAME = re.compile('^(?=.{1,80}$)[a-zA-Z0-9]([\\w,-]*[\\w])?$')
```
tomchop, line 34: This one is a bit harder to test, but

```suggestion
REGEX_SNAPSHOT_NAME = re.compile('^(?=.{1,80}$)[a-zA-Z0-9]([\\w,-]*\w)?$')
```

`\w` is just like `[a-zA-Z0-9]` ( plus `_`), maybe you can use that too?

libcloudforensics/providers/azure/internal/common.py
```diff
@@ -0,0 +1,132 @@
...
+  """
+  return ServicePrincipalCredentials(tenant=os.environ["AZURE_TENANT_ID"],
+                                     client_id=os.environ["AZURE_CLIENT_ID"],
+                                     secret=os.environ["AZURE_CLIENT_SECRET"])
+
+
+def ExecuteRequest(
+    client: Any,
+    func: str,
+    kwargs: Optional[Dict[str, str]] = None) -> List[Any]:
```
tomchop, line 55: Why not use `List[Dict]` here if that's what you're returning?

libcloudforensics/providers/azure/internal/common.py
```diff
@@ -0,0 +1,132 @@
...
+    if not next_link:
+      return responses
+
+
+def GenerateDiskName(snapshot: 'compute.AZSnapshot',
+                     disk_name_prefix: Optional[str] = None) -> str:
+  """Generate a new disk name for the disk to be created from the Snapshot.
+
+  The disk name must comply with the following RegEx:
+      - ^[\\w]{1-80}$
```
tomchop, line 91:
```suggestion
      - ^\w{1-80}$
```

libcloudforensics/providers/azure/internal/compute.py
```diff
@@ -0,0 +1,275 @@
...
+    if tags:
+      creation_data['tags'] = tags
+
+    try:
+      request = self.az_account.compute_client.snapshots.create_or_update(
+          self.resource_group_name,
+          snapshot_name,
+          creation_data)
+      while not request.done():
+        sleep(5)  # Wait 5 seconds before checking snapshot status again
```
tomchop, line 223: Same, do we wanna wait forever here?

libcloudforensics/providers/azure/internal/compute.py
```diff
@@ -0,0 +1,275 @@
...
+               region: str,
+               disk: AZDisk) -> None:
+    """Initialize the AZDisk class.
+
+    Args:
+      az_account (AZAccount): An Azure account object.
+      resource_id (str): The Azure ID of the snapshot.
+      name (str): The snapshot name.
+      region (str): The region in which the snapshot is located.
+      disk (AZDisk): The disk from which the snapshot was taken.
```
tomchop, line 256: Maybe rename this to `source_disk`?

libcloudforensics/providers/azure/internal/account.py
giovannt0: Tbh I couldn't find an actual reason; my local linter fails to import these for some reason. Seems to have happened to other people from the quick search I've done, but I didn't find any fix.

libcloudforensics/providers/azure/internal/account.py
giovannt0: `az` is the official command line to interact with Azure products

libcloudforensics/providers/azure/internal/account.py
giovannt0: What would be an appropriate timeout? For large disk operations things could take a little while

libcloudforensics/providers/azure/internal/common.py
giovannt0: While identical, having a double `\\` satisfies PEP 8 escape sequence (IDE complaining otherwise). I don't mind removing it and shutting off my PEP8 detector. Wdyt?

libcloudforensics/providers/azure/internal/common.py
giovannt0: We can't, as the Azure documentation specifies that the starting letter cannot be `_`, but only alphanumerical (which they define as being `[a-zA-Z0-9]`)
https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules

libcloudforensics/providers/azure/internal/common.py
giovannt0: Ah. remnants of copy-paste. I updated the docstring.

libcloudforensics/providers/azure/internal/compute.py
giovannt0: closing this comment (same question as above) and we can discuss on the other one for better visibility

libcloudforensics/providers/azure/internal/account.py
tomchop: I assume pylint complains about this, even though the libs are installed in its virtualenv? Could be that the underlying libs don't expose their members correctly. In that case, please add a note that even though the linters complain, the libs can be imported correctly

libcloudforensics/providers/azure/internal/account.py
tomchop: Hmm in that case it might not make that much sense to have a timeout as it will be hard to predict. As long as the user can Ctrl + C and bail out of waiting, we should be good. Feel free to disregard this comment.

libcloudforensics/providers/azure/internal/account.py
giovannt0: Correct! I added a note.

PR #175, opened 2020-06-26
Remove yapf config

giovannt0:
Yapf seems to be breaking our style guide more than fixing it.


giovannt0:
Following discussions, keeping this for now.

PR #179, opened 2020-06-29
Add unit tests for Azure snapshotting capabilities

giovannt0:
Add unit tests for #174

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ botocore
 azure-common
 azure-core
 azure-identity
-azure-mgmt-compute
+azure-mgmt-compute==13.0.0
 azure-mgmt-resource
 msrest
 msrestazure
--- a/tests/providers/azure/azure_test.py
+++ b/tests/providers/azure/azure_test.py
@@ -20,9 +20,13 @@ import mock
 from libcloudforensics.providers.azure.internal import account, compute, common
 from libcloudforensics.providers.azure import forensics
 
-FAKE_ACCOUNT = account.AZAccount(
-    'fake-subscription-id'
-)
+# pylint: disable=line-too-long
+with mock.patch('libcloudforensics.providers.azure.internal.common.GetCredentials') as mock_creds:
+  # pylint: enable=line-too-long
+  mock_creds.return_value = mock.Mock()
+  FAKE_ACCOUNT = account.AZAccount(
+      'fake-subscription-id'
+  )
 
 FAKE_INSTANCE = compute.AZVirtualMachine(
     FAKE_ACCOUNT,
@@ -316,6 +320,7 @@ class TestForensics(unittest.TestCase):
   """Test Azure forensics file."""
   # pylint: disable=line-too-long
 
+  @mock.patch('libcloudforensics.providers.azure.internal.common.GetCredentials')
   @mock.patch('libcloudforensics.providers.azure.internal.compute.AZSnapshot.Delete')
   @mock.patch('libcloudforensics.providers.azure.internal.compute.AZDisk.Snapshot')
   @mock.patch('libcloudforensics.providers.azure.internal.account.AZAccount.GetDisk')
@@ -329,7 +334,8 @@ class TestForensics(unittest.TestCase):
                           mock_get_boot_disk,
                           mock_get_disk,
                           mock_snapshot,
-                          mock_snapshot_delete):
+                          mock_snapshot_delete,
+                          mock_credentials):
     """Test that a disk copy is correctly created."""
     mock_create_disk.return_value.done.return_value = True
     mock_create_disk.return_value.result.return_value = MOCK_DISK_COPY
@@ -337,6 +343,7 @@ class TestForensics(unittest.TestCase):
     mock_get_boot_disk.return_value = FAKE_BOOT_DISK
     mock_snapshot.return_value = FAKE_SNAPSHOT
     mock_snapshot_delete.return_value = None
+    mock_credentials.return_value = mock.Mock()
 
     # CreateDiskCopy(zone, instance_name='fake-vm-name'). This should grab
     # the boot disk of the instance.
@@ -349,6 +356,7 @@ class TestForensics(unittest.TestCase):
     self.assertIsInstance(disk_copy, compute.AZDisk)
     self.assertEqual('fake_snapshot_name_f4c186ac_copy', disk_copy.name)
 
+  @mock.patch('libcloudforensics.providers.azure.internal.common.GetCredentials')
   @mock.patch('libcloudforensics.providers.azure.internal.compute.AZSnapshot.Delete')
   @mock.patch('libcloudforensics.providers.azure.internal.compute.AZDisk.Snapshot')
   @mock.patch('libcloudforensics.providers.azure.internal.account.AZAccount.GetDisk')
@@ -362,7 +370,8 @@ class TestForensics(unittest.TestCase):
                           mock_get_boot_disk,
                           mock_get_disk,
                           mock_snapshot,
-                          mock_snapshot_delete):
+                          mock_snapshot_delete,
+                          mock_credentials):
     """Test that a disk copy is correctly created."""
     mock_create_disk.return_value.done.return_value = True
     mock_create_disk.return_value.result.return_value = MOCK_DISK_COPY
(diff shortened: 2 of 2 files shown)


tests/providers/azure/azure_test.py
```diff
@@ -13,3 +13,410 @@
...
+    mock_request.assert_called_with(
+        mock.ANY,
+        'list_by_resource_group',
+        {'resource_group_name': disk.resource_group_name})
+
+  @mock.patch('libcloudforensics.providers.azure.internal.account.AZAccount.ListInstances')
+  @typing.no_type_check
+  def testGetInstance(self, mock_list_instances):
+    """Test that a particular instance from an account is retrieved."""
+    mock_list_instances.return_value = MOCK_LIST_INSTANCES
```
tomchop, line 164: No assert_called_with?

tests/providers/azure/azure_test.py
```diff
@@ -13,3 +13,410 @@
...
+        '/a/b/c/fake-resource-group/fake-vm-name', instance.resource_id)
+    self.assertEqual('fake-resource-group', instance.resource_group_name)
+    self.assertEqual('fake-region', instance.region)
+    self.assertEqual(['fake-zone'], instance.zones)
+
+  @mock.patch('libcloudforensics.providers.azure.internal.account.AZAccount.ListDisks')
+  @typing.no_type_check
+  def testGetDisk(self, mock_list_disks):
+    """Test that a particular disk from an account is retrieved."""
+    mock_list_disks.return_value = MOCK_LIST_DISKS
```
tomchop, line 177: No assert_called_with?

tests/providers/azure/azure_test.py
```diff
@@ -13,3 +13,410 @@
...
+    disk = FAKE_ACCOUNT.GetDisk('fake-disk-name')
+    self.assertEqual('fake-disk-name', disk.name)
+    self.assertEqual(
+        '/a/b/c/fake-resource-group/fake-disk-name', disk.resource_id)
+    self.assertEqual('fake-resource-group', disk.resource_group_name)
+    self.assertEqual('fake-region', disk.region)
+    self.assertEqual(['fake-zone'], disk.zones)
+
+  @mock.patch('azure.mgmt.compute.v2020_05_01.operations._disks_operations.DisksOperations.create_or_update')
```
tomchop, line 186: Oof, this sounds terrible. Are we going to have to change this mock whenever they update the library?

tests/providers/azure/azure_test.py
```diff
@@ -13,3 +13,410 @@
...
+    #     snapshot=FAKE_SNAPSHOT,
+    #     disk_name='new-forensics-disk',
+    #     disk_name_prefix='')
+    FAKE_ACCOUNT.CreateDiskFromSnapshot(
+        FAKE_SNAPSHOT, disk_name='new-forensics-disk')
+    mock_create_disk.assert_called_with(
+        FAKE_SNAPSHOT.resource_group_name,
+        'new-forensics-disk',
+        mock.ANY,
+        sku='Standard_LRS')
```
tomchop, line 215: Are we just testing that the calls don't raise any exceptions? There doesn't seem to be any assertions after this.

tests/providers/azure/azure_test.py
```diff
@@ -13,3 +13,410 @@
...
+  @typing.no_type_check
+  def testCreateDiskCopy1(self,
+                          mock_create_disk,
+                          mock_get_instance,
+                          mock_get_boot_disk,
+                          mock_get_disk,
+                          mock_snapshot,
+                          mock_snapshot_delete,
+                          mock_credentials):
+    """Test that a disk copy is correctly created."""
```
tomchop, line 339: I normally don't care too much about docstrings in tests, but since you have two very similar tests it would be nice to clarify the difference between both tests

requirements.txt
```diff
@@ -7,7 +7,7 @@ botocore
 azure-common
 azure-core
 azure-identity
-azure-mgmt-compute
+azure-mgmt-compute==13.0.0
```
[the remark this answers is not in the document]
giovannt0: If not, the github actions (the ones running tests through `setup.py` and the other ones running them through `nosetests`) download two different versions of the azure package, making the tests fail.

tests/providers/azure/azure_test.py
```diff
@@ -13,3 +13,410 @@
...
+    zones=['fake-zone']
+)
+MOCK_BOOT_DISK.name = 'fake-boot-disk-name'
+
+MOCK_DISK_COPY = mock.Mock(
+    id='/a/b/c/fake-resource-group/fake_snapshot_name_f4c186ac_copy',
+    location='fake-region',
+    zones=['fake-zone']
+)
+MOCK_DISK_COPY.name = 'fake_snapshot_name_f4c186ac_copy'
```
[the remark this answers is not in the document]
giovannt0: Because:

```
>>> mock = Mock(name='foo')
>>> mock.name
<Mock name='foo.name' id='4434041544'>

>>> mock = Mock()
>>> mock.name = 'foo'
>>> mock.name
'foo'
```

tests/providers/azure/azure_test.py
giovannt0: On what, `mock_list_instances `? It only takes an optional parameter which is not used here. We could add an `assert_called_once` though, if we want. Wdyt?

tests/providers/azure/azure_test.py
giovannt0: Same as above

tests/providers/azure/azure_test.py
giovannt0: Welcome to the delicious world of Azure's Python SDK :P 
Maybe there's some way to get whatever version is being used in code directly and then format the `mock.patch` statement accordingly? This would also allow us to un-pin the requirement. Either way, it'll be messy.

tests/providers/azure/azure_test.py
giovannt0: We're testing that the create disk function is called with the appropriate new disk name depending if we passed prefixes or whatnot. I did not add additional assertions on the resulting disk names as it would require us to change `mock_create_disk.return_value.result.return_value` values in-between tests which doesn't add much (the final disk name is being read from the disk created from the SDK)

tests/providers/azure/azure_test.py
giovannt0: The difference is clarified right before the `CreateDiskCopy` call in a comment. Should we move this comment to the docstrings for similar tests?

requirements.txt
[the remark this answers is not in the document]
tomchop: Hmm OK, but this should be investigated as it's really not the normal behavior (all packages should be downloaded from pypi, unless the pypi archive / wheel are out of sync)

tests/providers/azure/azure_test.py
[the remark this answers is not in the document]
tomchop: I was not expecting that. Can you add a comment (on this line only) saying that you need to manually assigne the attribute for it not to be a `Mock` object?

tests/providers/azure/azure_test.py
tomchop: That sounds good!

tests/providers/azure/azure_test.py
tomchop: Nah it's fine. But is this module intended to be accessed by library users? Otherwise it might be better to mock a higher-level function

tests/providers/azure/azure_test.py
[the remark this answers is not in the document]
giovannt0: Added!

tests/providers/azure/azure_test.py
giovannt0: I think so, since this is the actual function that the documentation asks you to call

giovannt0:
@tomchop updated, PTAL

tests/providers/azure/azure_test.py
tomchop: Yes, sorry I missed this.

tests/providers/azure/azure_test.py
tomchop: Ah I see it now! yeah, it might be better to put it in the docstring as this is also what shows up when the test fails.

giovannt0: Ack, done

PR #180, opened 2020-06-30
Add --start and --end to GCP log query functionality

### LINKED ISSUE #81: Add start/end time filter support to log acquisitions
- [x] AWS
- [ ] GCP

hacktobeer:
Will close #81

### DIFF
--- a/examples/gcp_cli.py
+++ b/examples/gcp_cli.py
@@ -69,8 +69,8 @@ def CreateDiskCopy(args: 'argparse.Namespace') -> None:
 
   disk = forensics.CreateDiskCopy(args.project,
                                   args.dst_project,
-                                  args.instance_name,
                                   args.zone,
+                                  instance_name=args.instance_name,
                                   disk_name=args.disk_name)
 
   print('Disk copy completed.')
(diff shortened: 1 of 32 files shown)


examples/gcp_cli.py
```diff
@@ -96,7 +96,18 @@ def QueryLogs(args: 'argparse.Namespace') -> None:
     args (argparse.Namespace): Arguments from ArgumentParser.
   """
   logs = gcp_log.GoogleCloudLog(args.project)
-  results = logs.ExecuteQuery(args.filter)
+
+  qfilter = ''
+  if args.start:
+    qfilter += 'timestamp>=\"{0:s}\" '.format(args.start)
```
giovannt0, line 102: You should be fine without escaping the double quote `'timestamp>="{0:s}" '`

examples/gcp_cli.py
```diff
@@ -96,7 +96,18 @@ def QueryLogs(args: 'argparse.Namespace') -> None:
...
+  if args.start and args.end:
+    qfilter += 'AND '
+  if args.end:
+    qfilter += 'timestamp<=\"{0:s}\" '.format(args.end)
```
giovannt0, line 106: Same as above

examples/gcp_cli.py
```diff
@@ -96,7 +96,18 @@ def QueryLogs(args: 'argparse.Namespace') -> None:
...
+  if args.filter:
+    qfilter += args.filter
```
tomchop, line 108: Don't you need a check to make sure that the two filters play well together? Like an `AND` inserted there somewhere?

examples/gcp_cli.py
```diff
@@ -96,7 +96,18 @@ def QueryLogs(args: 'argparse.Namespace') -> None:
     args (argparse.Namespace): Arguments from ArgumentParser.
   """
   logs = gcp_log.GoogleCloudLog(args.project)
-  results = logs.ExecuteQuery(args.filter)
+
+  qfilter = ''
+  if args.start:
```
tomchop, line 101: It might be nice to check that `--start` and `--end` are properly formatted and fail early if not.

examples/gcp_cli.py
hacktobeer: Line 103/104 contain the 'AND' logic.

examples/gcp_cli.py
hacktobeer: Done.

examples/gcp_cli.py
tomchop: From what I see `AND` is only inserted in between the two timestamps query, but not in between the timestamp query and the `filter` part. So if you specify `start` and `end` and `filter`, you'll end up with something like `start AND end filter` I would expect `start AND end AND filter`.

examples/gcp_cli.py
tomchop: If you're just gonna catch and raise the same error, maybe it would be worth displaying a message to the user explaining the error?

hacktobeer:
- The error raised currently looks like below, I thought adding anything would be too much:
"ValueError: time data 'asdada' does not match format '%Y-%m-%dT%H:%M:%SZ'".

- Added 'AND' filter buildup logic.

PR #186, opened 2020-06-30
Add missing Optional type hint where possible

### LINKED ISSUE #171: Be consistent in type hinting named parameters
Ideally we would have `def foo(bar: Optional[Type] = None) -> ReturnType` for named parameters which default to `None` instead of  `def foo(bar: Type = None) -> ReturnType`. As of now we have both across the codebase.

Note: we want to use `Optional` where it makes sense from `mypy` POV. `def foo(bar: Optional[str] = 'default-string') -> ReturnType` does not make sense because `Optional` essentially means `Union[type, None]`. In this case this would mean `Union[str, None]` and unless we're specifically allowing / expecting that `bar` may be `None`, `Optional` shouldn't be used here.

giovannt0:
Closes #171 

### DIFF
--- a/examples/aws_cli.py
+++ b/examples/aws_cli.py
@@ -149,6 +149,7 @@ def StartAnalysisVm(args: 'argparse.Namespace') -> None:
                                                             str(vm[1]),
                                                             vm[0].region))
 
+
 def ListImages(args: 'argparse.Namespace') -> None:
   """List AMI images and filter on AMI image 'name'.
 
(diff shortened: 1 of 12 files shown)


giovannt0:
@tomchop if you have the following:

```
def foo(arg1: Optional[int] = 5) -> int:  # arg1 is optional, therefore could be type hinted as such.
    return bar(arg1)  # mypy will complain here, because we're passing an Optional[int] to a method that's type hinted as 
                      # having its first parameter being an int, not an optional one.

def bar(arg1: int) -> int:
    return arg1
```

From what we tested with @sa3eed3ed, either we have not to type hint some parameters as `Optional` (which ends up confusing people reading through the codebase), or we have to use `--no-strict-optional`.

giovannt0:
As discussed: dropping commit `fa005f` that added more `Optional` annotations. Un-annotated parameters will stay this way to stay consistent with what `Optional` means to `mypy` rather to what it looks like it means from the user's perspective. This will be documented in upcoming doc release.
@sa3eed3ed @tomchop PTAL for the review.

PR #187, opened 2020-06-30
Remove unnecessary # type: ignore comments

### LINKED ISSUE #167: Limit the number of type hinting ignore statemetns
The codebase is riddled with `# type: ignore` statements, which defeats the purpose of using type hinting. Let's see if we can limit the usage of this.

giovannt0:
Closes #167 

### DIFF
--- a/examples/cli.py
+++ b/examples/cli.py
@@ -155,11 +155,12 @@ def Main() -> None:
   AddParser('gcp', gcp_subparsers, 'copydisk', 'Create a GCP disk copy.',
             args=[
                 ('dst_project', 'Destination GCP project.', ''),
-                ('instance_name', 'Name of the instance to copy disk from.',
-                 ''),
                 ('zone', 'Zone to create the disk in.', ''),
-                ('--disk_name', 'Name of the disk to copy. If None, the boot '
-                                'disk of the instance will be copied.', None)
+                ('--instance_name', 'Name of the instance to copy disk from.',
+                 ''),
+                ('--disk_name', 'Name of the disk to copy. If none specified, '
+                                'then --instance_name must be specified and '
+                                'the boot disk of the will be copied.', None)
             ])
   AddParser('gcp', gcp_subparsers, 'startvm', 'Start a forensic analysis VM.',
             args=[
--- a/examples/cli.py
+++ b/examples/cli.py
@@ -160,7 +160,8 @@ def Main() -> None:
                  ''),
                 ('--disk_name', 'Name of the disk to copy. If none specified, '
                                 'then --instance_name must be specified and '
-                                'the boot disk of the will be copied.', None)
+                                'the boot disk of the instance will be copied.',
+                 None)
             ])
   AddParser('gcp', gcp_subparsers, 'startvm', 'Start a forensic analysis VM.',
             args=[
--- a/libcloudforensics/providers/gcp/forensics.py
+++ b/libcloudforensics/providers/gcp/forensics.py
@@ -30,8 +30,8 @@ if TYPE_CHECKING:
 def CreateDiskCopy(
     src_proj: str,
     dst_proj: str,
-    instance_name: str,
     zone: str,
+    instance_name: Optional[str] = None,
     disk_name: Optional[str] = None,
     disk_type: str = 'pd-standard') -> 'compute.GoogleComputeDisk':
   """Creates a copy of a Google Compute Disk.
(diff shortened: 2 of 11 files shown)


hacktobeer:
Ran tests and some live e2e runs and works like a charm.

LGTM

examples/cli.py
```diff
@@ -155,11 +155,12 @@ def Main() -> None:
...
-                ('instance_name', 'Name of the instance to copy disk from.',
-                 ''),
                 ('zone', 'Zone to create the disk in.', ''),
-                ('--disk_name', 'Name of the disk to copy. If None, the boot '
-                                'disk of the instance will be copied.', None)
+                ('--instance_name', 'Name of the instance to copy disk from.',
+                 ''),
+                ('--disk_name', 'Name of the disk to copy. If none specified, '
+                                'then --instance_name must be specified and '
+                                'the boot disk of the will be copied.', None)
```
sa3eed3ed, line 163: the boot disk of the instance will be copied

libcloudforensics/providers/gcp/forensics.py
```diff
@@ -30,19 +30,19 @@
 def CreateDiskCopy(
     src_proj: str,
     dst_proj: str,
-    instance_name: str,
     zone: str,
+    instance_name: Optional[str] = None,
```
sa3eed3ed, line 34: why not ( : str = '' ) similar to other places in the code, I am not sure but I believe you are removing None and using empty objects to avoid explicit type comments or ignore statements

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -346,8 +346,8 @@ def GetOrCreateAnalysisVm(self,
 
   def ListInstanceByLabels(self,
                            labels_filter: Dict[str, str],
-                           filter_union: Optional[bool] = True
-                          )-> Dict[str, 'GoogleComputeInstance']:
+                           filter_union: bool = True
```
sa3eed3ed, line 349: in some places in the code you use Optional typing and in some other places you remove it, why is this not consistent?

libcloudforensics/providers/gcp/forensics.py
giovannt0: See #171 for context

libcloudforensics/providers/gcp/internal/compute.py
giovannt0: See #171 for context

sa3eed3ed: I am just thinking that this will be confusing for contributors.

I did a quick test and adding --no-strict-optional [here](https://github.com/google/cloud-forensics-utils/blob/50396979a6e3e330fedb186a4b5942ce9dc0cff3/.github/workflows/pylint_and_mypy.yml#L28) solved the problem. Or you thought about this already and it won't work for other reasons?

That way we can have Optional everywhere without ignore comments.

giovannt0: Ah great catch! Didn’t see this option in mypy. I will update PR #186 which was addressing this. I’ll then rebase this branch based on that.

giovannt0: Thanks @sa3eed3ed, I updated #186. Let's get that one merged in first.

libcloudforensics/providers/gcp/forensics.py
giovannt0: A bit more info on this one, as I realise my previous answer doesn't really fit your comment: prior to this PR `instance_name` was a required parameter, which makes little sense when you want to use `disk_name` which is optional (especially on the CLI, where it's cumbersome having to use the instance name + the disk name if you already know the latter to make your copy). To make it consistent with what we have with AWS (and soon Azure) and make the CLI a little more user friendly, I switched `instance_name` to optional. Then, why initialise it to `None` and not to `''` is just a personal preference on my side. If we want to assign it `''` then I would opt to create a separate issue to track this and go over the entire codebase to make it consistent with other optional parameters of type `string`.
Thoughts @tomchop @hacktobeer @sa3eed3ed ?

examples/cli.py
giovannt0: Oops, missed a word. Corrected.

libcloudforensics/providers/gcp/forensics.py
hacktobeer: My 2 cents is that it should be Optional if from a functional perspective the parameter *is* optional. So the typing should reflect the actual status of a parameter from the user perspective.

Setting a str to None or '' I am indifferent to. Afaik code flows (eg 'if str:')treat a str that is None or '' the same. As long as we use it consistently through the codebase.

sa3eed3ed: I agree that it is indifferent, I just realised that None is replaced with a '' in other places in the code as well and was wondering if there is a specific reason, I vote for consistency and I my subjective preference is None

giovannt0: #189

tomchop: FWIW I think it's better to use `None` than empty strings. Means that string functions like `len` or `split` will fail and bugs will be easier to debug.

PR #192, opened 2020-07-01
Default value for optional string parameters should be None

### LINKED ISSUE #189: Make named parameters of type string default to None and not ''
Make sure we default `string` named parameters to `None` across the codebase instead of `''` for consistency.

```
def foo(arg: str = None):
```

giovannt0:
Closes #189

### DIFF
--- a/libcloudforensics/providers/aws/forensics.py
+++ b/libcloudforensics/providers/aws/forensics.py
@@ -226,7 +226,7 @@ def StartAnalysisVm(
       raise RuntimeError('error - ListImages returns >1 AMI image: [{0:s}]'
                          .format(', '.join(image_names)))
     ami = ami_list[0]['ImageId']
-  assert ami  # Mypy: assert not None
+  assert ami  # Mypy: assert that ami is not None
 
   analysis_vm, created = aws_account.GetOrCreateAnalysisVm(
       vm_name,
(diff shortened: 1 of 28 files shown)


libcloudforensics/providers/aws/forensics.py
```diff
@@ -226,6 +226,7 @@ def StartAnalysisVm(
       raise RuntimeError('error - ListImages returns >1 AMI image: [{0:s}]'
                          .format(', '.join(image_names)))
     ami = ami_list[0]['ImageId']
+  assert ami  # Mypy: assert not None
```
sa3eed3ed, line 229: can you make the assert here and in other places more explicit.
i.e. "assert ami is not None"

libcloudforensics/providers/aws/forensics.py
```diff
@@ -164,7 +164,7 @@ def StartAnalysisVm(
     vm_name: str,
     default_availability_zone: str,
     boot_volume_size: int,
-    ami: str = '',
+    ami: Optional[str] = None,
```
tomchop, line 167: Doesn't `Optional[str]` default to `None`? e.g. do you need to specify `= None` each time? https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type

giovannt0: If you remove `= None` it seems that you are changing the argument to a non-default parameter and my IDE is not happy.

PR #196, opened 2020-07-03
Add GetBucketACLs and ListBucketObjects

Fryyyyy:
Plus tests, and CLI commands


examples/gcp_cli.py
```diff
@@ -163,3 +164,45 @@ def ListServices(args: 'argparse.Namespace') -> None:
...
+def GetBucketACLs(args: 'argparse.Namespace') -> None:
+  """Retrieve the Access Controls for a GCS bucket.
+
+  Args:
+    args (argparse.Namespace): Arguments from ArgumentParser.
+  """
+  gcs = gcp_storage.GoogleCloudStorage(args.project)
+  bucket_acls = gcs.GetBucketACLs(args.path)
+  for role in bucket_acls:
+    print('{}: {}'.format(role, ', '.join(bucket_acls[role])))
```
tomchop, line 178: Please adjust format string: `{0:s}` (see other sections of the code where this happens)

examples/gcp_cli.py
```diff
@@ -163,3 +164,45 @@ def ListServices(args: 'argparse.Namespace') -> None:
...
+
+def GetGCSObjectMetadata(args: 'argparse.Namespace') -> None:
+  """List the details of an object in a GCS bucket.
+
+  Args:
+    args (argparse.Namespace): Arguments from ArgumentParser.
+  """
+  gcs = gcp_storage.GoogleCloudStorage(args.project)
+  results = gcs.GetObjectMetadata(args.path)
+  if results.get('kind', '') == 'storage#objects':
```
tomchop, line 189: The default return value for `get` is `None`, so you can do the following:

```suggestion
  if results.get('kind') == 'storage#objects':
```

This also happens in other parts of the codebase.

examples/gcp_cli.py
```diff
@@ -163,3 +164,45 @@ def ListServices(args: 'argparse.Namespace') -> None:
...
+    for item in results.get('items', []):
+      for k, v in item.items():
```
tomchop, line 191: please use `key` and `value` here

examples/gcp_cli.py
```diff
@@ -163,3 +164,45 @@ def ListServices(args: 'argparse.Namespace') -> None:
...
+def ListBucketObjects(args: 'argparse.Namespace') -> None:
+  """List the objects in a GCS bucket.
+
+  Args:
+    args (argparse.Namespace): Arguments from ArgumentParser.
+  """
+  gcs = gcp_storage.GoogleCloudStorage(args.project)
+  results = gcs.ListBucketObjects(args.path)
+  for obj in results:
+    print('{}: {}'.format(obj.get('id', ''), obj.get('size', '?')))
```
tomchop, line 208: Please be a bit more explicit about the output values here in case they're not found

libcloudforensics/providers/gcp/internal/storage.py
```diff
@@ -83,10 +85,65 @@ def GetObjectMetadata(self,
...
+    if bucket.startswith('gs://'):
+      # Can change to removeprefix() in 3.9
+      bucket = bucket[5:]
+    gcs_bac = self.GcsApi().bucketAccessControls()
+    request = gcs_bac.list(
+        bucket=bucket, userProject=user_project)
+    # https://cloud.google.com/storage/docs/json_api/v1/bucketAccessControls#resource
+    ac_response = request.execute()
+    for item in ac_response.get('items', []):
+      if item.get('kind', '') == 'storage#bucketAccessControl':  # Sanity check
```
tomchop, line 123:
```suggestion
      if item.get('kind') == 'storage#bucketAccessControl':  # Sanity check
```

tests/providers/gcp/gcp_test.py
```diff
@@ -708,11 +741,35 @@ def testGetObjectMetadata(self, mock_gcs_api):
...
     self.assertEqual('MzFiYWIzY2M0MTJjNGMzNjUyZDMyNWFkYWMwODA5YTEgIGNvdW50MQo=', get_results['md5Hash'])
 
+  @typing.no_type_check
+  @mock.patch('libcloudforensics.providers.gcp.internal.storage.GoogleCloudStorage.GcsApi')
+  def testListBucketObjects(self, mock_gcs_api):
+    """Test GCS object List operation."""
+    api_list_object = mock_gcs_api.return_value.objects.return_value.list
+    api_list_object.return_value.execute.return_value = MOCK_GCS_BUCKET_OBJECTS
+    list_results = FAKE_GCS.ListBucketObjects('gs://fake-bucket')
+    self.assertEqual(len(list_results), 1)
```
giovannt0, line 756: Please format assertions as `self.assertEqual(expected, actual)`: `self.assertEqual(1, len(list_results))` so that it is consistent with the other tests.

Same comment for the assertions down below in `testGetBucketACLs`.

libcloudforensics/providers/gcp/internal/storage.py
```diff
@@ -14,7 +14,8 @@
 # limitations under the License.
 """Google Cloud Storage functionalities."""
 
-from typing import TYPE_CHECKING, Dict, Any, Optional, Tuple
+import collections
```
binglot, line 17: Since you're only using defaultdict(), wouldn't `from collections import defaultdict` be more appropriate?

libcloudforensics/providers/gcp/internal/storage.py
```diff
@@ -83,10 +85,65 @@ def GetObjectMetadata(self,
...
+    Args:
+      bucket (str):  Name of a bucket in GCS.
+
+    Returns:
+      List of Object Dicts (see GetObjectMetadata)
+    """
+    if bucket.startswith('gs://'):
+      # Can change to removeprefix() in 3.9
+      bucket = bucket[5:]
+    gcs_api_client = self.GcsApi().objects()
```
binglot, line 146: I find it a little convoluted that we have a class variable `self.gcs_api_client` which the function `GcsApi()` set and returns, and later we use the same name as a local variable but now we're calling `.objects()` on it. Perhaps something worth refactoring?

libcloudforensics/providers/gcp/internal/storage.py
Fryyyyy: This brings up a warning from the linter as it's against our internal style guide

libcloudforensics/providers/gcp/internal/storage.py
Fryyyyy: Agreed, and changed.

PR #197, opened 2020-07-03
create images from compressed archived GCE images (.tar.gz) in GCS

sa3eed3ed:
create images from compressed archived GCE images (.tar.gz) in GCS

### DIFF
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -513,7 +513,7 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
 
     Args:
       gcs_uri (str): Path to the compressed image archive
-          (image.tar.gz) in Cloud Storage. It must be gzip compressed
+          (image.tar.gz) in Cloud Storage. It must be a gzip compressed
           tar archive with the extension .tar.gz.
           ex: 'https://storage.cloud.google.com/foo/bar.tar.gz'
           'gs://foo/bar.tar.gz'
@@ -525,7 +525,7 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
       GoogleComputeImage: A Google Compute Image object.
 
     Raises:
-      ValueError: If the GCE Image name is invalid. If the extension of
+      ValueError: If the GCE Image name is invalid, or if the extension of
           the archived image is not valid.
     """
 
@@ -543,11 +543,11 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
     if ext != 'tar.gz':
       raise ValueError(
           'Image imported from {0:s} must be a GZIP compressed TAR '
-          'archive with the extension: .tar.gz'.format(
-              gcs_uri))
+          'archive with the extension: .tar.gz'.format(gcs_uri))
     gcs_uri = os.path.relpath(gcs_uri, 'gs://')
-    if not gcs_uri.startswith('https://storage.cloud.google.com'):
-      gcs_uri = os.path.join('https://storage.cloud.google.com', gcs_uri)
+    storage_link_url = 'https://storage.cloud.google.com'
+    if not gcs_uri.startswith(storage_link_url):
+      gcs_uri = os.path.join(storage_link_url, gcs_uri)
     image_body = {
         'name': name,
         "rawDisk": {
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -539,20 +539,18 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
       name = common.GenerateUniqueInstanceName('imported-image',
                                                common.COMPUTE_NAME_LIMIT)
 
-    ext = (os.path.split(gcs_uri)[-1]).partition('.')[-1].lower()
-    if ext != 'tar.gz':
+    if not gcs_uri.lower().endswith('.tar.gz'):
       raise ValueError(
           'Image imported from {0:s} must be a GZIP compressed TAR '
           'archive with the extension: .tar.gz'.format(gcs_uri))
     gcs_uri = os.path.relpath(gcs_uri, 'gs://')
-    storage_link_url = 'https://storage.cloud.google.com'
-    if not gcs_uri.startswith(storage_link_url):
-      gcs_uri = os.path.join(storage_link_url, gcs_uri)
+    if not gcs_uri.startswith(common.STORAGE_LINK_URL):
+      gcs_uri = os.path.join(common.STORAGE_LINK_URL, gcs_uri)
     image_body = {
         'name': name,
         "rawDisk": {
             'source': gcs_uri
-            }
+        }
     }
     gce_image_client = self.GceApi().images()
     request = gce_image_client.insert(
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -505,6 +505,62 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
     self.BlockOperation(response)
     return GoogleComputeImage(self.project_id, '', name)
 
+  def CreateImageFromGcsTarGz(
+      self,
+      gcs_uri: str,
+      name: Optional[str] = None) -> 'GoogleComputeImage':
+    """Creates a GCE image from a Gzip compressed Tar archive in GCS.
+
+    Args:
+      gcs_uri (str): Path to the compressed image archive
+          (image.tar.gz) in Cloud Storage. It must be gzip compressed
+          tar archive with the extension .tar.gz.
+          ex: 'https://storage.cloud.google.com/foo/bar.tar.gz'
+          'gs://foo/bar.tar.gz'
+          'foo/bar.tar.gz'
+      name (str): Optional. Name of the image to create. Default
+          is [src_disk.name]-[TIMESTAMP('%Y%m%d%H%M%S')].
+
+    Returns:
+      GoogleComputeImage: A Google Compute Image object.
+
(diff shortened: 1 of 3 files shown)


libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
...
 
+  def CreateImageFromGcsTarGz(
+      self,
+      gcs_uri: str,
+      name: Optional[str] = None) -> 'GoogleComputeImage':
+    """Creates a GCE image from a Gzip compressed Tar archive in GCS.
+
+    Args:
+      gcs_uri (str): Path to the compressed image archive
+          (image.tar.gz) in Cloud Storage. It must be gzip compressed
```
giovannt0, line 516: "It must be a gzip compressed tar archive, with .tar.gz. extension" ?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
...
+          'gs://foo/bar.tar.gz'
+          'foo/bar.tar.gz'
+      name (str): Optional. Name of the image to create. Default
+          is [src_disk.name]-[TIMESTAMP('%Y%m%d%H%M%S')].
+
+    Returns:
+      GoogleComputeImage: A Google Compute Image object.
+
+    Raises:
+      ValueError: If the GCE Image name is invalid. If the extension of
```
giovannt0, line 528: "If the GCE Image name is invalid, or if the extension of the archived image is invalid."?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
...
+    else:
+      name = common.GenerateUniqueInstanceName('imported-image',
+                                               common.COMPUTE_NAME_LIMIT)
+
+    ext = (os.path.split(gcs_uri)[-1]).partition('.')[-1].lower()
+    if ext != 'tar.gz':
+      raise ValueError(
+          'Image imported from {0:s} must be a GZIP compressed TAR '
+          'archive with the extension: .tar.gz'.format(
+              gcs_uri))
```
giovannt0, line 547: Doesn't this fit in one line? `'archive with the extension: .tar.gz'.format(gcs_uri))`

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
...
+    gcs_uri = os.path.relpath(gcs_uri, 'gs://')
+    if not gcs_uri.startswith('https://storage.cloud.google.com'):
```
giovannt0, line 549: While only used twice here, since we have most constants in `common.py`, maybe move the URL there?

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: yes that is true and this is the standard export generated by image export.

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: I thought about that but is might be confusing because all other functions took storage path as "gs://foo/bar" except this one takes the url link thus left it here

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: added "a"

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: done

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: yes, done

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: added that to a local variable though

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
     self.BlockOperation(response)
     return GoogleComputeImage(self.project_id, '', name)
 
+  def CreateImageFromGcsTarGz(
+      self,
+      gcs_uri: str,
+      name: Optional[str] = None) -> 'GoogleComputeImage':
```
tomchop, line 511: See my comment on https://github.com/google/cloud-forensics-utils/pull/192/files#r451488908, I'm not sure `= None` is needed here.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
...
+      if not common.REGEX_DISK_NAME.match(name):
+        raise ValueError(
+            'Image name {0:s} does not comply with {1:s}'.format(
+                name, common.REGEX_DISK_NAME.pattern))
+      name = name[:common.COMPUTE_NAME_LIMIT]
+    else:
+      name = common.GenerateUniqueInstanceName('imported-image',
+                                               common.COMPUTE_NAME_LIMIT)
+
+    ext = (os.path.split(gcs_uri)[-1]).partition('.')[-1].lower()
```
tomchop, line 542: Any reason there are `()` around `os.path.split`?

```suggestion
    ext = os.path.split(gcs_uri)[-1].partition('.')[-1].lower()
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
...
+    if ext != 'tar.gz':
```
tomchop, line 543: it's more straightforward to do `gcs_uri.lower().endswith('.tar.gz')`

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
...
+      raise ValueError(
+          'Image imported from {0:s} must be a GZIP compressed TAR '
+          'archive with the extension: .tar.gz'.format(gcs_uri))
+    gcs_uri = os.path.relpath(gcs_uri, 'gs://')
+    storage_link_url = 'https://storage.cloud.google.com'
```
tomchop, line 548: Maybe define `STORAGE_LINK_URL` this as a module constant?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -505,6 +505,62 @@ def CreateImageFromDisk(self,
...
+    if not gcs_uri.startswith(storage_link_url):
+      gcs_uri = os.path.join(storage_link_url, gcs_uri)
+    image_body = {
+        'name': name,
+        "rawDisk": {
+            'source': gcs_uri
+            }
+    }
+    gce_image_client = self.GceApi().images()
```
tomchop, line 557:
```suggestion
    image_body = {
        'name': name,
        "rawDisk": {
            'source': gcs_uri
        }
    }
    gce_image_client = self.GceApi().images()
```

tests/providers/gcp/gcp_test.py
```diff
@@ -581,6 +583,20 @@ def testListDisksByLabels(self, mock_gce_api, mock_labels):
...
+  @mock.patch('libcloudforensics.providers.gcp.internal.common.GoogleCloudComputeClient.GceApi')
+  def testCreateImageFromGcsTarGz(self, mock_gce_api, mock_gce_image, mock_block_operation):
+    """Test that images are correctly imported from compressed tar archives in GCS."""
+    mock_block_operation.return_value = None
+    mock_gce_image.return_value = FAKE_IMAGE
+    image_insert = mock_gce_api.return_value.images.return_value.insert
+    image_insert.return_value.execute.return_value = {'name': 'fake-image'}
+    image_object = FAKE_ANALYSIS_PROJECT.compute.CreateImageFromGcsTarGz(
+        'gs://fake-bucket/fake-folder/image.tar.gz', 'fake-image')
+    self.assertIn('fake-image', image_object.name)
```
tomchop, line 598: It would also be good to assert that `gce_image_client.insert` is called with the parameters you're expecting. You can use `image_insert.assert_called_with` for that

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: nothing, left over from the way I wanted to do this originally, will remove it

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: I will added to common module as suggested by @giovannt0, I thought it might be confusing, since most of the storage API methods takes object path as gs://foo/bar and not as a link url starting with https://storage.cloud.google.com, however it looks like it make sense for both of you

tests/providers/gcp/gcp_test.py
sa3eed3ed: done

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: Done

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: yes, changed it

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: looks like we will leave the None

PR #201, opened 2020-07-09
Implement cross-region and cross-account disk copy functionality

giovannt0:
Some changes following #174 

- Authentication is now more flexible. People can use environment variables or define their access keys in a `.json` file that is parsed. The library looks by default in `~/.azure/credentials.json` but it is also possible to pass an absolute path to the `AZURE_CREDENTIALS_PATH` environment variable. The advantage of the latter approach is that it replicates things you can do in AWS (define auth profiles).

- If the specified resource group name does not exists in the subscription, it will be automatically created.

- Disk copy functionality now allows you to select a different region for the copy to be created in, rather than using the region in which the source disk is.

- Disk copy functionality now allows you to make the copy in a different subscription within an Azure account.

- Disk copy functionality now allows you to make the copy in an entirely different Azure account:
  - In Azure, snapshots can be shared by generating a SAS (Shared Access Signature) downloadable link of the snapshot. Anyone with this link can download a `.vhd` copy of the snapshot.
  - As it is not possible to directly create a new disk out of a SAS link, the library generates (in the destination account) a temporary Azure Storage Account and corresponding Blob Container, and downloads the `.vhd` there.
  - As a third step, a new disk is created from the imported snapshot stored in the Azure Storage Account (in the destination account).
  - Things are cleaned up afterwards: all intermediary snapshots, storage account, SAS links.

Note that this complicated process is also necessary when creating a disk copy in a different region within the same account, as there is no other workaround at the moment.

The CLI tool, unit tests and end-to-end tests have all been updated to reflect these changes.

### DIFF
--- a/examples/az_cli.py
+++ b/examples/az_cli.py
@@ -30,7 +30,7 @@ def ListInstances(args: 'argparse.Namespace') -> None:
     args (argparse.Namespace): Arguments from ArgumentParser.
   """
 
-  az_account = account.AZAccount(args.subscription_id)
+  az_account = account.AZAccount(args.default_resource_group_name)
   instances = az_account.ListInstances(
       resource_group_name=args.resource_group_name)
 
@@ -47,7 +47,7 @@ def ListDisks(args: 'argparse.Namespace') -> None:
     args (argparse.Namespace): Arguments from ArgumentParser.
   """
 
-  az_account = account.AZAccount(args.subscription_id)
+  az_account = account.AZAccount(args.default_resource_group_name)
   disks = az_account.ListDisks(resource_group_name=args.resource_group_name)
 
   print('Disks found:')
@@ -62,8 +62,11 @@ def CreateDiskCopy(args: 'argparse.Namespace') -> None:
     args (argparse.Namespace): Arguments from ArgumentParser.
   """
   print('Starting disk copy...')
-  disk_copy = forensics.CreateDiskCopy(args.subscription_id,
-                                       args.instance_name,
-                                       args.disk_name,
-                                       args.disk_type)
+  disk_copy = forensics.CreateDiskCopy(args.default_resource_group_name,
+                                       instance_name=args.instance_name,
+                                       disk_name=args.disk_name,
+                                       disk_type=args.disk_type,
+                                       region=args.region,
+                                       src_profile=args.src_profile,
+                                       dst_profile=args.dst_profile)
   print('Done! Disk {0:s} successfully created.'.format(disk_copy.name))
--- a/examples/az_cli.py
+++ b/examples/az_cli.py
@@ -16,8 +16,12 @@
 
 from typing import TYPE_CHECKING
 
+from libcloudforensics import logging_utils
 from libcloudforensics.providers.azure.internal import account
 from libcloudforensics.providers.azure import forensics
+
+logging_utils.SetUpLogger(__name__)
+logger = logging_utils.GetLogger(__name__)
 
 if TYPE_CHECKING:
   import argparse
@@ -34,10 +38,10 @@ def ListInstances(args: 'argparse.Namespace') -> None:
   instances = az_account.ListInstances(
       resource_group_name=args.resource_group_name)
 
-  print('Instances found:')
+  logger.info('Instances found:')
   for instance in instances.values():
     boot_disk = instance.GetBootDisk().name
-    print('Name: {0:s}, Boot disk: {1:s}'.format(instance, boot_disk))
+    logger.info('Name: {0:s}, Boot disk: {1:s}'.format(instance, boot_disk))
 
 
 def ListDisks(args: 'argparse.Namespace') -> None:
@@ -50,9 +54,9 @@ def ListDisks(args: 'argparse.Namespace') -> None:
   az_account = account.AZAccount(args.default_resource_group_name)
   disks = az_account.ListDisks(resource_group_name=args.resource_group_name)
 
-  print('Disks found:')
+  logger.info('Disks found:')
   for disk in disks:
-    print('Name: {0:s}, Region: {1:s}'.format(disk, disks[disk].region))
+    logger.info('Name: {0:s}, Region: {1:s}'.format(disk, disks[disk].region))
 
 
 def CreateDiskCopy(args: 'argparse.Namespace') -> None:
@@ -61,7 +65,7 @@ def CreateDiskCopy(args: 'argparse.Namespace') -> None:
   Args:
     args (argparse.Namespace): Arguments from ArgumentParser.
   """
-  print('Starting disk copy...')
+  logger.info('Starting disk copy...')
   disk_copy = forensics.CreateDiskCopy(args.default_resource_group_name,
                                        instance_name=args.instance_name,
                                        disk_name=args.disk_name,
@@ -69,4 +73,4 @@ def CreateDiskCopy(args: 'argparse.Namespace') -> None:
                                        region=args.region,
                                        src_profile=args.src_profile,
                                        dst_profile=args.dst_profile)
-  print('Done! Disk {0:s} successfully created.'.format(disk_copy.name))
+  logger.info('Done! Disk {0:s} successfully created.'.format(disk_copy.name))
--- a/examples/cli.py
+++ b/examples/cli.py
@@ -159,7 +159,9 @@ def Main() -> None:
             ])
 
   # Azure parser options
-  az_parser.add_argument('subscription_id', help='The Azure subscription ID.')
+  az_parser.add_argument('default_resource_group_name',
+                         help='The default resource group name in which to '
+                              'create resources')
   az_subparsers = az_parser.add_subparsers()
   AddParser('az', az_subparsers, 'listinstances',
             'List instances in Azure subscription.',
@@ -182,8 +184,19 @@ def Main() -> None:
                                 'instance will be copied.', None),
                 ('--disk_type', 'The SKU name for the disk to create. '
                                 'Can be Standard_LRS, Premium_LRS, '
-                                'StandardSSD_LRS, or UltraSSD_LRS.',
-                 None)
+                                'StandardSSD_LRS, or UltraSSD_LRS.', None),
+                ('--region', 'The region in which to create the disk copy. If '
+                             'not provided, the disk copy will be created in '
+                             'the "eastus" region.', None),
+                ('--src_profile', 'The Azure profile information to use as '
+                                  'source account for the disk copy. Default '
+                                  'will look into environment variables to '
+                                  'authenticate the requests.', None),
+                ('--dst_profile', 'The Azure profile information to use as '
+                                  'destination account for the disk copy. If '
+                                  'not provided, the default behavior is to '
+                                  'use the same destination profile as the '
+                                  'source profile.', None)
             ])
 
   # GCP parser options
--- a/examples/cli.py
+++ b/examples/cli.py
@@ -184,10 +184,11 @@ def Main() -> None:
                                 'instance will be copied.', None),
                 ('--disk_type', 'The SKU name for the disk to create. '
                                 'Can be Standard_LRS, Premium_LRS, '
-                                'StandardSSD_LRS, or UltraSSD_LRS.', None),
+                                'StandardSSD_LRS, or UltraSSD_LRS. Default is '
+                                'Standard_LRS', 'Standard_LRS'),
                 ('--region', 'The region in which to create the disk copy. If '
                              'not provided, the disk copy will be created in '
-                             'the "eastus" region.', None),
+                             'the "eastus" region.', 'eastus'),
                 ('--src_profile', 'The Azure profile information to use as '
                                   'source account for the disk copy. Default '
                                   'will look into environment variables to '
--- a/libcloudforensics/providers/azure/forensics.py
+++ b/libcloudforensics/providers/azure/forensics.py
@@ -31,7 +31,7 @@ def CreateDiskCopy(
     instance_name: Optional[str] = None,
     disk_name: Optional[str] = None,
     disk_type: str = 'Standard_LRS',
-    region: Optional[str] = None,
+    region: str = 'eastus',
     src_profile: Optional[str] = None,
     dst_profile: Optional[str] = None) -> 'compute.AZDisk':
   """Creates a copy of an Azure Compute Disk.
@@ -48,8 +48,8 @@ def CreateDiskCopy(
     disk_type (str): Optional. The sku name for the disk to create. Can be
         Standard_LRS, Premium_LRS, StandardSSD_LRS, or UltraSSD_LRS. The
         default value is Standard_LRS.
-    region (str): Optional. The region in which to create the disk copy. If not
-        provided, the disk will be created in the default_region (eastus).
+    region (str): Optional. The region in which to create the disk copy.
+        Default is eastus.
     src_profile (str): Optional. The name of the source profile to use for the
         disk copy, i.e. the account information of the Azure account that holds
         the disk. For more information on profiles, see GetCredentials()
--- a/libcloudforensics/providers/azure/forensics.py
+++ b/libcloudforensics/providers/azure/forensics.py
@@ -100,6 +100,7 @@ def CreateDiskCopy(
     # which we import the previously created snapshot (cross-region/account
     # sharing).
     if diff_account or diff_region:
+      logger.info('Copy requested in a different destination account/region.')
       # Create a link to download the snapshot
       snapshot_uri = snapshot.GrantAccessAndGetURI()
       # Make a snapshot copy in the destination account from the link
--- a/libcloudforensics/providers/azure/forensics.py
+++ b/libcloudforensics/providers/azure/forensics.py
@@ -16,22 +16,25 @@
 
 from typing import TYPE_CHECKING, Optional
 
-from libcloudforensics.providers.azure.internal import account
-from libcloudforensics.providers.azure.internal import common
+from libcloudforensics.providers.azure.internal import account, common
 
 if TYPE_CHECKING:
   from libcloudforensics.providers.azure.internal import compute
 
 
 def CreateDiskCopy(
-    subscription_id: str,
+    resource_group_name: str,
     instance_name: Optional[str] = None,
     disk_name: Optional[str] = None,
-    disk_type: str = 'Standard_LRS') -> 'compute.AZDisk':
+    disk_type: str = 'Standard_LRS',
+    region: Optional[str] = None,
+    src_profile: Optional[str] = None,
+    dst_profile: Optional[str] = None) -> 'compute.AZDisk':
   """Creates a copy of an Azure Compute Disk.
 
   Args:
-    subscription_id (str): The Azure subscription ID to use.
+    resource_group_name (str): The resource group in which to create the disk
+        copy.
     instance_name (str): Optional. Instance name of the instance using the
         disk to be copied. If specified, the boot disk of the instance will be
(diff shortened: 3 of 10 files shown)


libcloudforensics/providers/azure/internal/account.py
```diff
@@ -217,3 +258,187 @@ def CreateDiskFromSnapshot(
...
+        account_url=storage_account_url, credential=storage_account_access_key)
+
+    # Create a container within the Storage to receive the imported snapshot
+    container_name = storage_account_name + '-container'
+    snapshot_vhd_name = snapshot.name + '.vhd'
+    container_client = blob_service_client.get_container_client(container_name)
+    try:
+      container_client.create_container()
+    except ResourceExistsError:
+      # The container already exists, so we can re-use it
```
tomchop, line 318: Issue a warning here?

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -217,3 +258,187 @@ def CreateDiskFromSnapshot(
...
+
+    Raises:
+      RuntimeError: If the disk could not be created.
+    """
+
+    if not region:
+      region = self.default_region
+
+    # Create a temporary Azure account storage to import the snapshot
+    storage_account_name = hashlib.sha1(
```
tomchop, line 302: Why use `sha1` instead of a random hex blob?

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -13,15 +13,20 @@
...
 # Pylint complains about the import but the library imports just fine,
 # so we can ignore the warning.
 # pylint: disable=import-error
+from azure.core.exceptions import ResourceExistsError
 from azure.mgmt.compute import ComputeManagementClient
 from azure.mgmt.compute.v2016_04_30_preview.models import DiskCreateOption
+from azure.mgmt.resource import SubscriptionClient
+from azure.mgmt.storage import StorageManagementClient
+from azure.storage.blob import BlobServiceClient
+from azure.mgmt.resource import ResourceManagementClient
```
tomchop, line 29: Please import modules or packages, not classes. https://google.github.io/styleguide/pyguide.html#22-imports

libcloudforensics/providers/azure/forensics.py
```diff
@@ -16,22 +16,25 @@
...
   from libcloudforensics.providers.azure.internal import compute
 
 
 def CreateDiskCopy(
-    subscription_id: str,
+    resource_group_name: str,
     instance_name: Optional[str] = None,
     disk_name: Optional[str] = None,
-    disk_type: str = 'Standard_LRS') -> 'compute.AZDisk':
+    disk_type: str = 'Standard_LRS',
```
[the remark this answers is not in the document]
giovannt0: It is indeed. This is not the default behavior with other providers, though. This disk type value is basically the "cheap" option. I think making it default to the same disk type as the source disk is a good idea; however I would opt to have this in a separate PR. Opened #210.

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -217,3 +258,187 @@ def CreateDiskFromSnapshot(
...
+        storage_account_name,
+        creation_data
+    )
+    storage_account = request.result()
+    storage_account_keys = self.storage_client.storage_accounts.list_keys(
+        self.default_resource_group_name, storage_account_name)
+    storage_account_keys = {v.key_name: v.value
+                            for v in storage_account_keys.keys}
+    storage_account_id = storage_account.id  # type: str
+    storage_account_key = storage_account_keys['key1']  # type: str
```
tomchop, line 413: What's `key1` here?

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -217,3 +258,187 @@ def CreateDiskFromSnapshot(
...
+    request = self.storage_client.storage_accounts.create(
+        self.default_resource_group_name,
+        storage_account_name,
+        creation_data
+    )
+    storage_account = request.result()
+    storage_account_keys = self.storage_client.storage_accounts.list_keys(
+        self.default_resource_group_name, storage_account_name)
+    storage_account_keys = {v.key_name: v.value
+                            for v in storage_account_keys.keys}
```
tomchop, line 411: Shouldn't `v` be `k`? You're accessing keys it seems, not values. Also, using `key` as a variable name would be better.

```suggestion
                            for key in storage_account_keys.keys}
```

libcloudforensics/providers/azure/forensics.py
```diff
@@ -54,18 +70,47 @@ def CreateDiskCopy(
...
+      # Make a snapshot copy in the destination account from the link
+      new_disk = dst_account.CreateDiskFromSnapshotURI(
+          snapshot,
+          snapshot_uri,
+          disk_name_prefix=common.DEFAULT_DISK_COPY_PREFIX,
+          disk_type=disk_type)
+      # Revoke download link and delete the initial copy
+      snapshot.RevokeAccessURI()
+      snapshot.Delete()
+      return new_disk
```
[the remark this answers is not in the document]
giovannt0: Sounds good, opted for the `else:` block

libcloudforensics/providers/azure/internal/common.py
```diff
@@ -34,21 +35,103 @@
...
+    except ValueError as exception:
+      raise ValueError('Could not decode JSON file. Please verify the file '
+                       'format: {0:s}'.format(str(exception)))
+    if not account_info:
+      raise ValueError(
+          'Profile name {0:s} not found in credentials file {1:s}'.format(
+              profile_name, path))
+    required_entries = ['subscriptionId', 'clientId', 'clientSecret',
+                        'tenantId']
+    if not all(entry in account_info for entry in required_entries):
```
tomchop, line 123: I assume all of these must have values, right? In that case, you can do 
```suggestion
    if not all(account_info.get(entry) for entry in required_entries):
```

libcloudforensics/providers/azure/internal/common.py
```diff
@@ -80,7 +163,7 @@ def ExecuteRequest(
     request = getattr(client, func)
     response = request(**kwargs)
     responses.append(response)
-    next_link = response.next_link
+    next_link = response.next_link if hasattr(response, 'next_link') else None
```
tomchop, line 163: See `getattr`: https://docs.python.org/3/library/functions.html#getattr and its default return value
```suggestion
    next_link = response.getattr('next_link', None)
```

libcloudforensics/providers/azure/internal/compute.py
```diff
@@ -275,3 +275,20 @@ def Delete(self) -> None:
...
       raise RuntimeError('Could not delete snapshot {0:s}: {1:s}'
                          .format(self.resource_id, str(exception)))
+
+  def GrantAccessAndGetURI(self) -> str:
+    """Grant access to a snapshot and return its access URI.
+
+    Returns:
+      str: The access URI for the snapshot.
+    """
+    snapshot = self.az_account.compute_client.snapshots.grant_access(
```
tomchop, line 285: Is this really a `snapshot`? Or a `grant_access` operation?

libcloudforensics/providers/azure/internal/account.py
giovannt0: No particular reason, just wanted something that was linked to the particular snapshot (but there again, no requirement to have that). Let me know if you want this changed to a random hex blob.

giovannt0:
Thanks @tomchop, updated; PTAL

tests/providers/azure/azure_test.py
```diff
@@ -238,6 +257,77 @@ def testCreateDiskFromSnapshot(self, mock_create_disk):
...
+    # pylint: disable=protected-access
+    mock_create.return_value.result.return_value = MOCK_STORAGE_ACCOUNT
+    mock_list_keys.return_value = MOCK_LIST_KEYS
+    account_id, account_key = FAKE_ACCOUNT._CreateStorageAccount('fakename')
+    self.assertEqual('fakestorageid', account_id)
+    self.assertEqual('fake-key-value', account_key)
+
+    with self.assertRaises(ValueError):
+      _, _ = FAKE_ACCOUNT._CreateStorageAccount(
+          'fake-non-conform-name')
```
tomchop, line 328: You can also check that the error string matches what you expect (since the error is pretty broad)

```suggestion
    with self.assertRaises(ValueError) as error:
      _, _ = FAKE_ACCOUNT._CreateStorageAccount(
          'fake-non-conform-name')
    # str(error.exception) will contain the exception string. You can access it *outside* the with statement
```

libcloudforensics/providers/azure/internal/account.py
giovannt0: Taken from https://docs.microsoft.com/en-us/samples/azure-samples/storage-python-manage/storage-python-manage/.
When creating a storage account, there're two keys associated to it, with `key1` being the one required to access the storage account (https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python)

libcloudforensics/providers/azure/internal/common.py
giovannt0: Correct, changed

libcloudforensics/providers/azure/internal/compute.py
giovannt0: Ah, yeah, that's a pretty bad name. Updated to `access_request`

tests/providers/azure/azure_test.py
giovannt0: Added in all statements, makes it clearer to the reader

giovannt0:
Added a bunch of logging everywhere with the new upstream logger brought by #204 

```
libcloudforensics az <resource_group> copydisk --disk_name <disk_name> --region westus
Starting disk copy...
[2020-07-20 21:47:08,785] [libcloudforensics.providers.azure.forensics] INFO     Disk copy of <disk_name> started...
[2020-07-20 21:47:08,786] [libcloudforensics.providers.azure.internal.compute] INFO     Creating snapshot: <disk_name>_snapshot
[2020-07-20 21:47:57,798] [libcloudforensics.providers.azure.internal.compute] INFO     Snapshot <disk_name>_snapshot successfully created
[2020-07-20 21:47:58,834] [libcloudforensics.providers.azure.forensics] INFO     Copy requested in a different destination account/region.
[2020-07-20 21:47:58,834] [libcloudforensics.providers.azure.internal.compute] INFO     Generating SAS URI for snapshot: <disk_name>_snapshot
[2020-07-20 21:48:29,843] [libcloudforensics.providers.azure.internal.compute] INFO     SAS URI generated: https://md-vxbs[TRUNCATED]
[2020-07-20 21:48:29,843] [libcloudforensics.providers.azure.internal.account] INFO     Creating storage account: 28bee920529b4[TRUNCATED]
[2020-07-20 21:48:34,793] [libcloudforensics.providers.azure.internal.account] INFO     Storage account 28bee920529b4[TRUNCATED] successfully created
[2020-07-20 21:48:54,156] [libcloudforensics.providers.azure.internal.account] INFO     Creating blob container 28bee920529b4[TRUNCATED]-container
[2020-07-20 21:48:55,068] [libcloudforensics.providers.azure.internal.account] INFO     Blob container 28bee920529b4[TRUNCATED]-container successfully created
[2020-07-20 21:48:55,069] [libcloudforensics.providers.azure.internal.account] INFO     Importing snapshot to container from URI https://md-vxbs[TRUNCATED]. Depending on the size of the snapshot, this process is going to take a while.
[2020-07-20 21:49:06,508] [libcloudforensics.providers.azure.internal.account] INFO     Snapshot successfully imported from URI https://md-vxbs[TRUNCATED]
[2020-07-20 21:49:06,509] [libcloudforensics.providers.azure.internal.account] INFO     Creating disk: evidence_<disk_name>_snapshot_d654fe89_copy
[2020-07-20 21:49:14,517] [libcloudforensics.providers.azure.internal.account] INFO     Disk evidence_<disk_name>_snapshot_d654fe89_copy successfully created
[2020-07-20 21:49:14,517] [libcloudforensics.providers.azure.internal.account] INFO     Deleting storage account: 28bee920529b4[TRUNCATED]
[2020-07-20 21:49:18,415] [libcloudforensics.providers.azure.internal.account] INFO     Storage account 28bee920529b4[TRUNCATED] successfully deleted
[2020-07-20 21:49:18,418] [libcloudforensics.providers.azure.internal.compute] INFO     Revoking SAS URI for snapshot <disk_name>_snapshot
[2020-07-20 21:49:49,874] [libcloudforensics.providers.azure.internal.compute] INFO     SAS URI revoked for snapshot <disk_name>_snapshot
[2020-07-20 21:49:49,874] [libcloudforensics.providers.azure.internal.compute] INFO     Deleting snapshot: <disk_name>_snapshot
[2020-07-20 21:50:25,375] [libcloudforensics.providers.azure.internal.compute] INFO     Snapshot <disk_name>_snapshot successfully deleted.
[2020-07-20 21:50:25,375] [libcloudforensics.providers.azure.forensics] INFO     Disk <disk_name> successfully copied to evidence_<disk_name>_snapshot_d654fe89_copy
Done! Disk evidence_<disk_name>_snapshot_d654fe89_copy successfully created.
```

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -37,16 +44,44 @@ class AZAccount:
...
+          operations. For more information on profiles, see GetCredentials()
+          in libcloudforensics.providers.azure.internal.common.py. Default
+          does not use profiles and will authenticate to Azure using
+          environment variables.
     """
-    self.subscription_id = subscription_id
-    self.credentials = common.GetCredentials()
-    self.compute_client = ComputeManagementClient(
+    self.subscription_id, self.credentials = common.GetCredentials(profile_name)
+    self.default_region = default_region or 'eastus'
```
hacktobeer, line 66: the 'eastus' default is buried pretty deep here, is there no way to make that clearer in the calling signature of the function on some level?

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -217,3 +262,213 @@ def CreateDiskFromSnapshot(
...
+      logger.warning('Reusing existing container: {0:s}'.format(container_name))
+
+    # Download the snapshot from the URI to the storage
+    copied_blob = blob_service_client.get_blob_client(
+        container_name, snapshot_vhd_name)
+    logger.info('Importing snapshot to container from URI {0:s}. '
+                'Depending on the size of the snapshot, this process is going '
+                'to take a while.'.format(snapshot_uri))
+    copied_blob.start_copy_from_url(snapshot_uri)
+    while copied_blob.get_blob_properties().copy.status != 'success':
```
hacktobeer, line 335: I don't know what other status can be returned, but if it returns eg 'error' or something this will loop and loop and loop.

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -217,3 +262,213 @@ def CreateDiskFromSnapshot(
...
+          'Storage account name {0:s} does not comply with {1:s}'.format(
+              storage_account_name, common.REGEX_ACCOUNT_STORAGE_NAME.pattern))
+
+    if not region:
+      region = self.default_region
+
+    creation_data = {
+        'location': region,
+        'sku': {
+            'name': 'Standard_RAGRS'
```
hacktobeer, line 408: Add a comment so we get some context on this default.
https://docs.microsoft.com/en-us/rest/api/storagerp/srp_sku_types

examples/az_cli.py
```diff
@@ -47,7 +47,7 @@ def ListDisks(args: 'argparse.Namespace') -> None:
     args (argparse.Namespace): Arguments from ArgumentParser.
   """
 
-  az_account = account.AZAccount(args.subscription_id)
+  az_account = account.AZAccount(args.default_resource_group_name)
   disks = az_account.ListDisks(resource_group_name=args.resource_group_name)
 
   print('Disks found:')
```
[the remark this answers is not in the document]
giovannt0: Agreed, updated. We should change as well the `print` statements in the other CLI files. Opened #212

examples/cli.py
```diff
@@ -182,8 +184,19 @@ def Main() -> None:
                                 'instance will be copied.', None),
                 ('--disk_type', 'The SKU name for the disk to create. '
                                 'Can be Standard_LRS, Premium_LRS, '
-                                'StandardSSD_LRS, or UltraSSD_LRS.',
-                 None)
+                                'StandardSSD_LRS, or UltraSSD_LRS.', None),
+                ('--region', 'The region in which to create the disk copy. If '
+                             'not provided, the disk copy will be created in '
+                             'the "eastus" region.', None),
```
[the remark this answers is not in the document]
giovannt0: We do not, the `AZAccount` class will take care of that.

examples/cli.py
```diff
@@ -182,8 +184,19 @@ def Main() -> None:
                                 'instance will be copied.', None),
                 ('--disk_type', 'The SKU name for the disk to create. '
                                 'Can be Standard_LRS, Premium_LRS, '
-                                'StandardSSD_LRS, or UltraSSD_LRS.',
-                 None)
+                                'StandardSSD_LRS, or UltraSSD_LRS.', None),
```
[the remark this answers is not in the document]
giovannt0: We do in this one! Thanks, good catch

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -13,19 +13,26 @@
...
 # Pylint complains about the import but the library imports just fine,
 # so we can ignore the warning.
 # pylint: disable=import-error
-from azure.mgmt.compute import ComputeManagementClient
-from azure.mgmt.compute.v2016_04_30_preview.models import DiskCreateOption
-from msrestazure.azure_exceptions import CloudError
+from azure.core import exceptions
+from azure.mgmt import compute as azure_compute
+from azure.mgmt import resource, storage
+from azure.mgmt.compute.v2020_05_01 import models
```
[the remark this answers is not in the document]
giovannt0: Yes :( Not much we can do here.

examples/cli.py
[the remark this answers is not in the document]
giovannt0: Actually, I updated this to default to 'eastus' given your next comment

libcloudforensics/providers/azure/internal/account.py
giovannt0: Updated as default parameter value for more visibility

libcloudforensics/providers/azure/internal/account.py
giovannt0: Yeah, we pretty much have this problem with every Azure call. The thing is we have no way to predict how long these operations will take (since the size of the disk to copy will greatly influence the time spent looping there). We talked with @tomchop in one of the previous PRs about having a `MAX_RETRY` control variable, but we ended up not using it (for the reason mentioned previously, i.e. we can't decide on a good value for this variable without knowing the use case). Users can just quit the program if the feel like it's hanging. The call to the logger should provide them with at least a sense of "things are happening on Azure, let's wait"

libcloudforensics/providers/azure/internal/account.py
giovannt0: Added

libcloudforensics/providers/azure/internal/account.py
hacktobeer: It would be pending when busy but can't we raise an error with the description when the status is 'failed'?
See https://docs.microsoft.com/en-us/rest/api/storageservices/get-blob-properties

giovannt0: I see, added safe guard

PR #203, opened 2020-07-10
Add authentication through access key for AWS

### LINKED ISSUE #202: Support other sources of credentials for boto3
Hi,

Right now, `libcloudforensics` only supports credentials read from the `~/.aws/config` file.

Would you consider adding support for other sources ?

[This page](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials) lists the configuration possibilities for boto3, which includes passing credentials as parameters when creating a Session object :
```python
import boto3

session = boto3.Session(
    aws_access_key_id=ACCESS_KEY,
    aws_secret_access_key=SECRET_KEY,
    aws_session_token=SESSION_TOKEN
)
```

Looking at the code from `account.py` ([here](https://github.com/google/cloud-forensics-utils/blob/master/libcloudforensics/providers/aws/internal/account.py#L77)), it seems it would only require minor changes (for example an optional switch to use credentials read from CLI/when using the library) to configure boto3 this way.

This would make `libclouforensics` usable without AWS CLI and without a configuration file, which stores credentials on disk.

This could be useful when integrating `libcloudforensics` in other projects, which might handle tokens and keys without using the CLI (using [get_session_token](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts.html#STS.Client.get_session_token) from boto3 for example).

Regards,

giovannt0:
Closes #202 

### DIFF
--- a/libcloudforensics/providers/aws/internal/account.py
+++ b/libcloudforensics/providers/aws/internal/account.py
@@ -43,7 +43,10 @@ class AWSAccount:
 
   def __init__(self,
                default_availability_zone: str,
-               aws_profile: Optional[str] = None) -> None:
+               aws_profile: Optional[str] = None,
+               aws_access_key_id: Optional[str] = None,
+               aws_secret_access_key: Optional[str] = None,
+               aws_session_token: Optional[str] = None) -> None:
     """Initialize the AWS account.
 
     Args:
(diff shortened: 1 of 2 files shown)


libcloudforensics/providers/aws/internal/account.py
```diff
@@ -43,14 +43,26 @@ class AWSAccount:
...
 
     Args:
       default_availability_zone (str): Default zone within the region to create
           new resources in.
       aws_profile (str): Optional. The AWS profile defined in the AWS
           credentials file to use.
+      aws_access_key_id (str): Optional. If provided together with
+          aws_secret_access_key and aws_session_token, authenticate to AWS
+          using these parameters instead of the credential file.
+      aws_secret_access_key (str): Optional.  If provided together with
```
tomchop, line 60:
```suggestion
      aws_secret_access_key (str): Optional. If provided together with
```

libcloudforensics/providers/aws/internal/account.py
```diff
@@ -43,14 +43,26 @@ class AWSAccount:
...
+          aws_access_key_id and aws_session_token, authenticate to AWS
+          using these parameters instead of the credential file.
+      aws_session_token (str): Optional.   If provided together with
```
tomchop, line 63:
```suggestion
      aws_session_token (str): Optional. If provided together with
```

PR #204, opened 2020-07-10
Add custom logger and add more logging

### LINKED ISSUE #61: More verbose logging

giovannt0:
- Add a custom logger with fancy colours (thanks @tomchop)
- Add more log messages in the forensics modules
- Disable un-wanted (ResourceWarning and UserWarning) warning messages in e2e tests to keep things clean (we really don't care about unclosed ssl socket while the test's running)

```
Launching unittests with arguments python -m unittest aws_e2e.EndToEndTest.testEncryptedVolumeCopy in ~/cloud-forensics-utils/tests/providers/aws
[2020-07-10 18:19:09,698] [libcloudforensics.providers.aws.forensics] INFO     No AMI provided, fetching one for Ubuntu 18.04
[2020-07-10 18:19:10,996] [libcloudforensics.providers.aws.forensics] INFO     Starting analysis VM new-vm-for-analysis
[2020-07-10 18:21:46,386] [libcloudforensics.providers.aws.forensics] INFO     VM started.
[2020-07-10 18:21:46,386] [libcloudforensics.providers.aws.forensics] INFO     VM ready.
[2020-07-10 18:21:47,365] [libcloudforensics.providers.aws.forensics] INFO     Volume copy of vol-xxx started...
[2020-07-10 18:22:18,453] [libcloudforensics.providers.aws.forensics] INFO     Created snapshot: snap-yyy
[2020-07-10 18:22:36,248] [libcloudforensics.providers.aws.forensics] INFO     Volume vol-xxx successfully copied to vol-zzz
[2020-07-10 18:22:36,248] [libcloudforensics.providers.aws.forensics] INFO     Cleaning up...
[2020-07-10 18:22:36,905] [libcloudforensics.providers.aws.forensics] INFO     Done
[2020-07-10 18:23:08,812] [aws_e2e             ] INFO     Deleting volume: vol-zzz.
[2020-07-10 18:23:25,052] [aws_e2e             ] INFO     Volume vol-zzz successfully deleted.
```

Closes #61 


libcloudforensics/logging_utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+  Returns:
+    logging.Logger: A logger.
+  """
+  logger = logging.getLogger(name)
+  logger.setLevel(logging.INFO)
+  console_handler = logging.StreamHandler(sys.stdout)
+  formatter = Formatter(random_color=True)
+  console_handler.setFormatter(formatter)
+  logger.addHandler(console_handler)
```
tomchop, line 119: This will add a new console_handler every time you call `GetLogger`. For each handler added, a log line will be displayed when you call the logging functions... so you might want to rename that function to SetUp logger or something, and use `logging.getLogger(name)` directly in other modules that need to log things.

libcloudforensics/providers/aws/forensics.py
```diff
@@ -111,16 +114,20 @@ def CreateVolumeCopy(zone: str,
...
         volume_to_copy.volume_id))
     snapshot = volume_to_copy.Snapshot()
+    logger.info('Created snapshot: {0:s}'.format(snapshot.snapshot_id))
 
     source_account_id = source_account.GetAccountInformation('Account')
     destination_account_id = destination_account.GetAccountInformation(
         'Account')
 
     if source_account_id != destination_account_id:
+      logger.info('External account detected')
```
tomchop, line 127: Maybe also specify which account is the external account?

libcloudforensics/providers/gcp/forensics.py
```diff
@@ -69,13 +72,13 @@ def CreateDiskCopy(
       instance = src_project.compute.GetInstance(instance_name)
       disk_to_copy = instance.GetBootDisk()
 
-    common.LOGGER.info('Disk copy of {0:s} started...'.format(
+    logger.info('Disk copy of {0:s} started...'.format(
         disk_to_copy.name))
     snapshot = disk_to_copy.Snapshot()
     new_disk = dst_project.compute.CreateDiskFromSnapshot(
         snapshot, disk_name_prefix='evidence', disk_type=disk_type)
     snapshot.Delete()
```
tomchop, line 80: You can also add the intermediary steps here (snapshot created, disk created from snapshot, snapshot deleted), maybe as `.debug()` statements if you want to be able to use the log level to chose what to display and what not.

libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -25,18 +24,19 @@
 from google.auth import default
 from google.auth.exceptions import DefaultCredentialsError, RefreshError
 from googleapiclient.discovery import build
+from libcloudforensics import logging_utils  # pylint: disable=ungrouped-imports
```
tomchop, line 27: Why are we disabling the ungrouped imports here? Also, see my other comment about just using the native `logging.getLogger()` function

libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -142,7 +142,7 @@ def CreateService(service_name: str,
           cache_discovery=False)
       service_built = True
     except socket.timeout:
-      LOGGER.info(
+      logger.info(
```
tomchop, line 145: Change to warning?

tests/providers/aws/aws_e2e.py
```diff
@@ -15,14 +15,29 @@
...
+
+@typing.no_type_check
+def IgnoreWarnings(test_func):
+  """Disable logging of warning messages."""
+  def DoTest(self, *args, **kwargs):
+    with warnings.catch_warnings():
+      warnings.simplefilter("ignore", ResourceWarning)
+      warnings.simplefilter("ignore", UserWarning)
+      test_func(self, *args, **kwargs)
+  return DoTest
```
tomchop, line 39: Why do we do this?

libcloudforensics/providers/gcp/internal/common.py
giovannt0: pylint sees the fake import under `if TYPE_CHECKING` so it wants the actual import there as well

tests/providers/aws/aws_e2e.py
giovannt0: This makes logging very ugly because it just floods the console with resource warnings about unclosed ssl sockets etc.
Thus it's really hard to see the logged messages which carry the information that you actually care about.

libcloudforensics/logging_utils.py
giovannt0: but you need to add a console handler for each new logger being created no? E.g. if calling two times getLogger but with different names, don't you need to both times create and add the console handler? In the meantime, I added a safeguard so that if the same logger is queried, no new handler is created.

tomchop: Correct. But if for some reason `GetLogger` is called twice with the same name, since loggers are singletons, handlers will be added twice. The check you added should be enough to not have this problem; but I'd still rename the method to `SetupLogger` since you're doing more than just fetching the singleton.

tests/providers/aws/aws_e2e.py
tomchop: Fair enough. Can you expand the docstring to add a bit more detail about what warnings we're trying to suppress?

giovannt0: Done

libcloudforensics/logging_utils.py
giovannt0: Added

PR #205, opened 2020-07-13
Adding GKE APIs and Cluster Get Operation

sa3eed3ed:
Adding GKE APIs and Cluster Get Operation, more functions in following PRs

*created constants in gcp_test.py as class constants instead of global constants, there are too many module global constants used only in one class and this is becoming difficult to read, these should all be moved to class constants for better readability, if you agree we can open a new issue for that


libcloudforensics/providers/gcp/internal/gke.py
```diff
@@ -0,0 +1,75 @@
...
+if TYPE_CHECKING:
+  import googleapiclient
+
+class GoogleKubernetesEngine:
+  """Class to call Google Kubernetes Engine (GKE) APIs.
+
+  https://cloud.google.com/kubernetes-engine/docs/reference/rest
+
+  Attributes:
+    gke_api_client: Client to interact with GKE APIs.
```
giovannt0, line 29: Maybe specify the types of the attributes in the docstring?

libcloudforensics/providers/gcp/internal/gke.py
```diff
@@ -0,0 +1,75 @@
...
+      return self.gke_api_client
+    self.gke_api_client = common.CreateService(
+        'container', self.GKE_API_VERSION)
+    return self.gke_api_client
+
+  def GetCluster(self, name: str) -> Dict[str, Any]:
+    """ Gets the details of a specific cluster.
+
+    Args:
+      name (string): The name (project, location, cluster) of the cluster to retrieve.
```
giovannt0, line 63: I think in other parts of the codebase we use `str` instead of `string` but this is being really picky, feel free to ignore

libcloudforensics/providers/gcp/internal/gke.py
```diff
@@ -0,0 +1,76 @@
...
+
+  def __init__(self, project_id: Optional[str] = None) -> None:
+    """Initialize the GoogleCloudStorage object.
+
+    Args:
+      project_id (str): Optional. Google Cloud project ID.
+    """
+
+    self.gke_api_client = None
+    self.project_id = project_id
```
tomchop, line 43: `project_id` doesn't seem to be used anywhere?

tests/providers/gcp/gcp_test.py
```diff
@@ -824,6 +825,26 @@ def testBlockOperation(self, mock_gcb_api):
...
+  }
+
+  @typing.no_type_check
+  @mock.patch('libcloudforensics.providers.gcp.internal.gke.GoogleKubernetesEngine.GkeApi')
+  def testGetCluster(self, mock_gke_api):
+    """Test GKE cluster Get operation."""
+    api_cluster_object = mock_gke_api.return_value.projects.return_value.locations.return_value.clusters.return_value
+    api_cluster_object.get.return_value.execute.return_value = GoogleKubernetesEngineTest.MOCK_GKE_CLUSTER_OBJECT
+    get_results = GoogleKubernetesEngineTest.FAKE_GKE.GetCluster('projects/fake-project/locations/fake-region/clusters/fake-cluster')
+    self.assertEqual(GoogleKubernetesEngineTest.MOCK_GKE_CLUSTER_OBJECT, get_results)
```
tomchop, line 845: I'm fine with the first two statements here be left on one line, but the last two can be broken down into many shorter lines easily enough. Can you please move the `pylint` statement to here to scope it down a bit?

libcloudforensics/providers/gcp/internal/gke.py
sa3eed3ed: it is not used, just added it by convention as for other services, will probably need it in future functions. Will remove it for now and can add it later

tests/providers/gcp/gcp_test.py
sa3eed3ed: moved the lint statement before the mock.patch line to avoid complains, when I break mock.patch lines they don't look necessarily readable for me, but I can break it if you think it looks better, I can also do the same for all other tests.

tomchop: Yeah that's exactly what I meant by "you can leave the first two lines as is", I think the mock stuff will look less readable if we attempt to split it. My comment was aimed at the other two lines which seemed to be able to fit if doing something like

```python
    self.assertEqual(
        GoogleKubernetesEngineTest.MOCK_GKE_CLUSTER_OBJECT,
        get_results)
```

and

```python
    get_results = GoogleKubernetesEngineTest.FAKE_GKE.GetCluster(
        'projects/fake-project/locations/fake-region/clusters/fake-cluster')
```

PR #215, opened 2020-07-22
Add Azure VM functionality

giovannt0:
As for GCP & AWS

### DIFF
--- a/libcloudforensics/providers/azure/forensics.py
+++ b/libcloudforensics/providers/azure/forensics.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """Forensics on Azure."""
 
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, List, Dict, Tuple
 
 from libcloudforensics import logging_utils
 from libcloudforensics.providers.azure.internal import account, common
@@ -125,3 +125,66 @@ def CreateDiskCopy(
     raise RuntimeError(error_msg)
 
   return new_disk
+
+
+def StartAnalysisVm(
+    resource_group_name: str,
+    vm_name: str,
+    boot_disk_size: int,
+    ssh_public_key: str,
+    cpu_cores: int = 4,
+    memory_in_mb: int = 8192,
+    region: str = 'eastus',
+    attach_disks: Optional[List[str]] = None,
+    tags: Optional[Dict[str, str]] = None,
+    dst_profile: Optional[str] = None
+    ) -> Tuple['compute.AZVirtualMachine', bool]:
+  """Start a virtual machine for analysis purposes.
+
+  Look for an existing Azure virtual machine with name vm_name. If found,
+  this instance will be started and used as analysis VM. If not found, then a
+  new vm with that name will be created, started and returned. Note that if a
+  new vm is created, you should provide the ssh_public_key parameter.
+
+  Args:
+    resource_group_name (str): The resource group in which to create the
+        analysis vm.
+    vm_name (str): The name for the virtual machine.
+    boot_disk_size (int): The size of the analysis VM boot disk (in GB).
+    ssh_public_key (str): A SSH public key data to associate with the
+        VM. This must be provided as otherwise the VM will not be
+        accessible.
+    cpu_cores (int): Number of CPU cores for the analysis VM.
+    memory_in_mb (int): The memory size (in MB) for the analysis VM.
+    region (str): Optional. The region in which to create the VM.
+        Default is eastus.
+    attach_disks (List[str]): Optional. List of disk names to attach to the VM.
+    tags (Dict[str, str]): Optional. A dictionary of tags to add to the
+        instance, for example {'TicketID': 'xxx'}. An entry for the instance
+        name is added by default.
+    dst_profile (str): The name of the destination profile to use for the vm
+        creation, i.e. the account information of the Azure account in which
+        to create the vm. For more information on profiles,
+        see GetCredentials() in
+        libcloudforensics.providers.azure.internal.common.py
+
+  Returns:
+    Tuple[AZVirtualMachine, bool]: a tuple with a virtual machine object
+        and a boolean indicating if the virtual machine was created or not.
+  """
+
+  az_account = account.AZAccount(resource_group_name,
+                                 default_region=region,
+                                 profile_name=dst_profile)
+
+  analysis_vm, created = az_account.GetOrCreateAnalysisVm(
+      vm_name,
+      boot_disk_size,
+      cpu_cores,
+      memory_in_mb,
+      ssh_public_key,
+      tags=tags)
+
+  for disk_name in (attach_disks or []):
+    analysis_vm.AttachDisk(az_account.GetDisk(disk_name))
+  return analysis_vm, created
--- a/libcloudforensics/providers/azure/forensics.py
+++ b/libcloudforensics/providers/azure/forensics.py
@@ -151,9 +151,9 @@ def StartAnalysisVm(
         analysis vm.
     vm_name (str): The name for the virtual machine.
     boot_disk_size (int): The size of the analysis VM boot disk (in GB).
-    ssh_public_key (str): A SSH public key data to associate with the
-        VM. This must be provided as otherwise the VM will not be
-        accessible.
+    ssh_public_key (str): A SSH public key data (OpenSSH format) to associate
+        with the VM, e.g. ssh-rsa AAAAB3NzaC1y... This must be provided as
+        otherwise the VM will not be accessible.
     cpu_cores (int): Number of CPU cores for the analysis VM.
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -13,3 +13,5 @@ azure-mgmt-storage==11.1.0
 azure-storage-blob
 msrest
 msrestazure
+sshpubkeys
+pycryptodome
```
giovannt0, line 17: This requirement (`pycryptodome`) is used by the CLI tool (see PR #217)

libcloudforensics/providers/azure/forensics.py
```diff
@@ -125,3 +125,66 @@ def CreateDiskCopy(
...
+  new vm is created, you should provide the ssh_public_key parameter.
+
+  Args:
+    resource_group_name (str): The resource group in which to create the
+        analysis vm.
+    vm_name (str): The name for the virtual machine.
+    boot_disk_size (int): The size of the analysis VM boot disk (in GB).
+    ssh_public_key (str): A SSH public key data to associate with the
+        VM. This must be provided as otherwise the VM will not be
+        accessible.
```
tomchop, line 156: Is this a path to the key? the b64 blob?

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -381,6 +387,314 @@ def CreateDiskFromSnapshotURI(
...
+                    'ssh': {
+                        'publicKeys': [{
+                            'path': '/home/AzureUser/.ssh/authorized_keys',
+                            'keyData': ssh_public_key}]
+                    }
+                }
+            },
+            'networkProfile': {
+                'networkInterfaces': [{'id': self._CreateNetworkInterfaceForVM(
+                    vm_name, region)}]
```
tomchop, line 487:
```suggestion
                'networkInterfaces': [
                    {'id': self._CreateNetworkInterfaceForVM(vm_name, region)}
                ]
```

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -381,6 +387,314 @@ def CreateDiskFromSnapshotURI(
...
+
+    Returns:
+      str: The id of the created network interface.
+
+    Raises:
+      ValueError: if vm_name is not provided.
+      RuntimeError: If no network interface could be created.
+    """
+    if not vm_name:
+      raise ValueError('vm_name cannot be None.')
```
tomchop, line 585:
```suggestion
    if not vm_name:
      raise ValueError('vm_name must be specified. Provided: {0!s}'.format(vm_name))
```

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -381,6 +387,314 @@ def CreateDiskFromSnapshotURI(
...
+
+    network_interface_name = '{0:s}-nic'.format(vm_name)
+    ip_config_name = '{0:s}-ipconfig'.format(vm_name)
+
+    # Check if the network interface already exists, and returns its ID if so.
+    try:
+      nic = self.network_client.network_interfaces.get(
+          self.default_resource_group_name, network_interface_name)
+      nic_id = nic.id  # type: str
+      return nic_id
```
tomchop, line 598: Maybe we can `return nic.id` directly?

requirements.txt
tomchop: By add it here and not in PR #217 then?

giovannt0: Because I did not think about doing this :D Removed it, will add it again in the other PR.

libcloudforensics/providers/azure/forensics.py
giovannt0: The actual OpenSSH formatted key `ssh-rsa AAAAB3Nz...dxxEZWU=`
Added indication in the docstring

libcloudforensics/providers/azure/internal/account.py
giovannt0: This would break mypy:

```
libcloudforensics/providers/azure/internal/account.py:601: error: Returning Any from function declared to return "str"
```

Not that you cannot specify the type on the return line either:
```
libcloudforensics/providers/azure/internal/account.py:601: error: misplaced type annotation
```

libcloudforensics/providers/azure/forensics.py
```diff
@@ -125,3 +125,66 @@ def CreateDiskCopy(
...
+  Look for an existing Azure virtual machine with name vm_name. If found,
+  this instance will be started and used as analysis VM. If not found, then a
+  new vm with that name will be created, started and returned. Note that if a
+  new vm is created, you should provide the ssh_public_key parameter.
+
+  Args:
+    resource_group_name (str): The resource group in which to create the
+        analysis vm.
+    vm_name (str): The name for the virtual machine.
+    boot_disk_size (int): The size of the analysis VM boot disk (in GB).
```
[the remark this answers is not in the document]
giovannt0: We don't use this notation with other providers; `memory_in_mb` was named that way so it would match the official Azure notation (https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinesizes/list). If this is bothering us I'd rather rename `memory_in_mb` to just `memory` and specify its unit in the docstring only. Wdyt?

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -381,6 +387,316 @@ def CreateDiskFromSnapshotURI(
...
+                    }
+                }
+            },
+            'networkProfile': {
+                'networkInterfaces': [
+                    {'id': self._CreateNetworkInterfaceForVM(vm_name, region)}
+                ]
+            }
+        }
+    }  # type: Dict[str, Any]
```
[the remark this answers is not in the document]
giovannt0: We have a little of everything in here: we have `Dict[str, str]`, but then we have nested dictionary structures up to four levels, `Dict[str, Dict[str, Dict[str, Dict[str, Any]]]]` while some of these inner dictionaries are `Dict[str, List[Dict[str, str]]`. 

Therefore for simplicity, I annotated all of this `Dict[str, Any]`

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -381,6 +387,316 @@ def CreateDiskFromSnapshotURI(
...
+      region = self.default_region
+
+    network_interface_name = '{0:s}-nic'.format(vm_name)
+    ip_config_name = '{0:s}-ipconfig'.format(vm_name)
+
+    # Check if the network interface already exists, and returns its ID if so.
+    try:
+      nic = self.network_client.network_interfaces.get(
+          self.default_resource_group_name, network_interface_name)
+      nic_id = nic.id  # type: str
```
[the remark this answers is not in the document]
giovannt0: This would break mypy:

```
libcloudforensics/providers/azure/internal/account.py:601: error: Returning Any from function declared to return "str"
```

Not that you cannot specify the type on the return line either:
```
libcloudforensics/providers/azure/internal/account.py:601: error: misplaced type annotation
```

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -381,6 +387,316 @@ def CreateDiskFromSnapshotURI(
...
+      request = self.network_client.network_interfaces.create_or_update(
+          self.default_resource_group_name,
+          network_interface_name,
+          creation_data)
+      request.wait()
+    except azure_exceptions.CloudError as exception:
+      raise RuntimeError('Could not create network interface: {0:s}'.format(
+          str(exception)))
+
+    network_interface_id = request.result().id  # type: str
```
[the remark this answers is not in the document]
giovannt0: See above answer

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -381,6 +387,316 @@ def CreateDiskFromSnapshotURI(
...
+    return network_interface_id
+
+  def _CreateNetworkInterfaceElements(
+      self,
+      vm_name: str,
+      region: Optional[str] = None) -> Tuple[Any, ...]:
```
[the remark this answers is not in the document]
giovannt0: This is a shortened notation for `Tuple[Any, Any, Any]` since all elements in that tuple are of the same type. You can use the dots `...` whenever you want to specify a type that repeats `n` times

giovannt0:
Thanks @binglot, PTAL

libcloudforensics/providers/azure/forensics.py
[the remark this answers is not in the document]
binglot: I agree with you and would be happy to leave it, just made sure that it was a conscious choice :) Great work Theo!

libcloudforensics/providers/azure/internal/account.py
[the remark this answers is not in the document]
binglot: Oh I didn't know that - thanks, that's something I just learnt then :)

PR #217, opened 2020-07-22
Add CLI Azure VM functionality

giovannt0:
Marking as draft until #215 is in.

### DIFF
--- a/examples/az_cli.py
+++ b/examples/az_cli.py
@@ -13,8 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Demo CLI tool for Azure."""
-
+import os
 from typing import TYPE_CHECKING
+from Crypto.PublicKey import RSA
 
 from libcloudforensics import logging_utils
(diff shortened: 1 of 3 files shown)


examples/az_cli.py
```diff
@@ -74,3 +75,81 @@ def CreateDiskCopy(args: 'argparse.Namespace') -> None:
                                        src_profile=args.src_profile,
                                        dst_profile=args.dst_profile)
   logger.info('Done! Disk {0:s} successfully created.'.format(disk_copy.name))
+
+
+def StartAnalysisVm(args: 'argparse.Namespace') -> None:
```
sa3eed3ed, line 80: It is better to have the vm object returned here, allows for further operations and checking the creation status without intializing a vm using its name. It might be a problem if the creation failed because  the user must first verify the VM exist before using it to avoid problems unhandled creation errors. when I look to the gcp one, it does return:
Returns:
    Tuple(GoogleComputeInstance, bool): A tuple with a virtual machine object
        and a boolean indicating if the virtual machine was created or not.

PR #218, opened 2020-07-22
Replace print statements by call to logger in CLI tool

### LINKED ISSUE #212: Replace print statements in CLI files by calls to the logger
Use #204 in the various CLI files instead of the print statements.

giovannt0:
Closes #212 

### DIFF
--- a/tools/aws_cli.py
+++ b/tools/aws_cli.py
@@ -26,6 +26,10 @@ from typing import TYPE_CHECKING
 from libcloudforensics.providers.aws.internal import account
 from libcloudforensics.providers.aws.internal import log as aws_log
 from libcloudforensics.providers.aws import forensics
+from libcloudforensics import logging_utils
+
+logging_utils.SetUpLogger(__name__)
+logger = logging_utils.GetLogger(__name__)
 
 if TYPE_CHECKING:
   import argparse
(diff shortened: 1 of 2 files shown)


examples/aws_cli.py
```diff
@@ -117,7 +121,7 @@ def StartAnalysisVm(args: 'argparse.Namespace') -> None:
     args (argparse.Namespace): Arguments from ArgumentParser.
   """
   if args.attach_volumes and len(args.attach_volumes.split(',')) > 11:
-    print('error: --attach_volumes must be < 11')
+    logger.error('error: --attach_volumes must be < 11')
```
sa3eed3ed, line 124: wouldn't be there an Error automatically printed by the logger before the message, shouldn't the 'error:' be removed, it looks to me like it will be: ERROR  error: --atta...

examples/aws_cli.py
```diff
@@ -126,7 +130,8 @@ def StartAnalysisVm(args: 'argparse.Namespace') -> None:
     # Check if volumes parameter exists and if there
     # are any empty entries.
     if not (volumes and all(elements for elements in volumes)):
-      print('error: parameter --attach_volumes: {0:s}'.format(args.attach_volumes))
+      logger.error('error: parameter --attach_volumes: {0:s}'.format(
```
sa3eed3ed, line 133: same regarding the additional error: here and in other places

PR #221, opened 2020-07-23
Add support for volume type selection in AWS

### LINKED ISSUE #220: Add disk type option for CreateVolumeCopy in AWS
Allow users to choose the volume type when making a volume copy, and default to the source volume type if none choses.

giovannt0:
This PR:

- Adds `--volume_type` CLI argument for AWS, allowing users to select the volume type they want to have for their volume copies

- Adds  `--boot_volume_type` CLI argument for AWS, allowing users to select the volume type they want to have for their analysis VM

The default behavior for volume copies is to use the same volume type as the source volume.

Closes #220 

### DIFF
--- a/libcloudforensics/providers/aws/internal/account.py
+++ b/libcloudforensics/providers/aws/internal/account.py
@@ -420,6 +420,7 @@ class AWSAccount:
       snapshot: ebs.AWSSnapshot,
       volume_name: Optional[str] = None,
       volume_name_prefix: Optional[str] = None,
+      volume_type: str = 'gp2',
       kms_key_id: Optional[str] = None,
       tags: Optional[Dict[str, str]] = None) -> ebs.AWSVolume:
     """Create a new volume based on a snapshot.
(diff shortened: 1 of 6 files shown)


libcloudforensics/providers/aws/internal/account.py
```diff
@@ -428,6 +429,9 @@ def CreateVolumeFromSnapshot(
       snapshot (AWSSnapshot): Snapshot to use.
       volume_name (str): Optional. String to use as new volume name.
       volume_name_prefix (str): Optional. String to prefix the volume name with.
+      volume_type (str): Optional. The volume type for the volume to create.
+          Can be one of 'standard'|'io1'|'gp2'|'sc1'|'st1'. The default is
+          'gp2'.
```
tomchop, line 434: Should we raise an exception if the volume type is not one of these?

giovannt0: Good idea, done

PR #223, opened 2020-07-29
Add custom error class

giovannt0:
Closes #71 

Introduces new error classes:

- CredentialsConfigurationError
- InvalidFileFormatError
- InvalidNameError
- ResourceNotFoundError
- ResourceCreationError
- ResourceDeletionError

Tests updated accordingly.

Signed-off-by: Theo Giovanna <gtheo@google.com>

PR #225, opened 2020-07-29
Refactor AWS module

### LINKED ISSUE #222: Refactor AWS internal module
Move methods like ListImages, ListVolumes, etc. to `ec2.py` to match closer the organization that we have under GCP. Make use of `@property` decorator in `account.py`

giovannt0:
Closes #222. Together with #234, this PR harmonizes the structure under which cloud provider functionality are implemented, following GCP's module organisation. `account.py` now only contains references to modules (ec2, ebs, kms, etc). 

This breaks down the code a little more and separate functionality better.

Unit tests, end to end tests and CLI files are updated as well to use the new structure.


libcloudforensics/providers/aws/internal/ebs.py
```diff
@@ -305,3 +305,308 @@ def ShareWithAWSAccount(self, aws_account_id: str) -> None:
         },
         OperationType='add'
     )
+
+
+class AWSEBS:
```
tomchop, line 310: Might be clearer to call this `AwsEbs`, or `AwsEBS`? Same with other class names.

tomchop:
Assigning @Fryyyyy as second reviewer for this.

libcloudforensics/providers/aws/internal/ebs.py
giovannt0: @tomchop Isn't it against the style guide to not keep case for acronyms such as AWS? Back when we introduced AWS (see [here](https://github.com/google/cloud-forensics-utils/pull/28/commits/16c57242c0c4ddf9a9215688aaa1c5ccd7ca0623#r404878599)) we decided to go with uppercase. If we're fine with deviating from the style guide I'm happy to change all impacted classes (FWIW, I also find it not so readable as it is now)

tomchop: Yeah but to acronyms that follow each other are hard to read. There's probably a rule about this somewhere but I can't fin it. Maybe we should go with what makes the most sense?

giovannt0: Agreed, happy to push the changes. I'll update similar class names to `AwsXXX`

Fryyyyy: Another option: just EBS. It's already under the AWS provider...

libcloudforensics/providers/aws/internal/ebs.py
```diff
@@ -305,3 +305,308 @@ def ShareWithAWSAccount(self, aws_account_id: str) -> None:
...
+  """Class that represents AWS EBS storage services."""
+
+  def __init__(self,
+               aws_account: 'account.AWSAccount') -> None:
+    """Initialize the AWS ebs client object.
```
Fryyyyy, line 315:
```suggestion
    """Initialize the AWS EBS client object.
```

libcloudforensics/providers/aws/internal/ebs.py
```diff
@@ -305,3 +305,308 @@ def ShareWithAWSAccount(self, aws_account_id: str) -> None:
...
+      str: The information requested.
+
+    Raises:
+      KeyError: If the requested information doesn't exist.
+    """
+    account_information = self.aws_account.ClientApi(
+        common.ACCOUNT_SERVICE).get_caller_identity()  # type: Dict[str, str]
+    if not account_information.get(info):
+      raise KeyError('Key must be one of ["UserId", "Account", "Arn"]')
+    return account_information[info]
```
Fryyyyy, line 573: I'm not sure how/where this is called but could we not just return all of them? It's only 3 values, could just let the callee filter if necessary and if someone wanted all the info they would currently have to call this 3 times.

libcloudforensics/providers/aws/internal/ebs.py
```diff
@@ -305,3 +305,308 @@ def ShareWithAWSAccount(self, aws_account_id: str) -> None:
...
+      ValueError: If the volume name does not comply with the RegEx.
+    """
+
+    # Max length of tag values in AWS is 255 characters
+    user_id = self.GetAccountInformation('UserId')
+    volume_id = user_id + snapshot.volume.volume_id
+    volume_id_crc32 = '{0:08x}'.format(
+        binascii.crc32(volume_id.encode()) & 0xffffffff)
+    truncate_at = 255 - len(volume_id_crc32) - len('-copy') - 1
+    assert snapshot.name  # Mypy: assert that snapshot.name is not None
```
Fryyyyy, line 597: I think these should be `is not None` globally.
Firstly it means the function could also raise an AssertionError but also Assertions are normally used for debugging and can be disabled globally based on runtime or environment variables

libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -137,3 +142,368 @@ def AttachVolume(self,
...
+          object will be used.
+
+    Returns:
+      List[AWSInstance]: A list of EC2 Instance objects. If no instance with
+          matching name tag is found, the method returns an empty list.
+    """
+
+    matching_instances = []
+    instances = self.ListInstances(region=region)
+    for instance_id in instances:
```
Fryyyyy, line 274: `matching_instances = [x for x in instances if x: x.name == instance_name]` if we're feeling fancy

Fryyyyy: Or do we not allow list comprehensions? I forget

libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -137,3 +142,368 @@ def AttachVolume(self,
...
+      RuntimeError: If the virtual machine cannot be found or created.
+    """
+
+    # Re-use instance if it already exists, or create a new one.
+    try:
+      instances = self.GetInstancesByName(vm_name)
+      if instances:
+        created = False
+        return instances[0], created
+    except RuntimeError:
```
Fryyyyy, line 380: We desperately need more specific error types here

libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -137,3 +142,368 @@ def AttachVolume(self,
...
+            'AvailabilityZone': self.aws_account.default_availability_zone}
+    }
+    if ssh_key_name:
+      vm_args['KeyName'] = ssh_key_name
+    # Create the instance in AWS
+    try:
+      instance = client.run_instances(**vm_args)
+      # If the call to run_instances was successful, then the API response
+      # contains the instance ID for the new instance.
+      instance_id = instance['Instances'][0]['InstanceId']
```
Fryyyyy, line 419: Could also throw IndexError/KeyError?

libcloudforensics/providers/aws/internal/kms.py
```diff
@@ -0,0 +1,124 @@
...
+  # the following cyclic import, as it it only used for type hints
+  from libcloudforensics.providers.aws.internal import account  # pylint: disable=cyclic-import
+
+
+class AWSKMS:
+  """Class that represents AWS KMS services."""
+
+  def __init__(self,
+               aws_account: 'account.AWSAccount') -> None:
+    """Initialize the AWS ec2 client object.
```
Fryyyyy, line 32:
```suggestion
    """Initialize the AWS KMS client object.
```

libcloudforensics/providers/aws/internal/ebs.py
giovannt0: Replaced by: 
```
if not snapshot.name:
  snapshot.name = snapshot.snapshot_id
```
In practice this method is never called with `snapshot.name = None`, this is only to satisfy `mypy`.

libcloudforensics/providers/aws/internal/ec2.py
giovannt0: List comprehensions is cool; updated

libcloudforensics/providers/aws/internal/ec2.py
giovannt0: This will be taken care of in #223

libcloudforensics/providers/aws/internal/ec2.py
giovannt0: If the call is successful it's expected to have the correct structure. If the call isn't we rely on the `boto3` library to throw an exception for whatever reason it couldn't do its job

giovannt0:
Thanks @tomchop , @Fryyyyy, PTAL

libcloudforensics/providers/aws/internal/ebs.py
tomchop: I agree, maybe it would be better to return a dict with those three keys and let the caller sort out which ones they need.

libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -137,3 +142,364 @@ def AttachVolume(self,
...
+          object will be used.
+
+    Returns:
+      List[AWSInstance]: A list of EC2 Instance objects. If no instance with
+          matching name tag is found, the method returns an empty list.
+    """
+
+    instances = self.ListInstances(region=region)
+    return [instance for _, instance in instances.items() if
+            instance.name == instance_name]
```
tomchop, line 274: use `values()` instead of `items()` if you don't care about the keys
```suggestion
    return [instance for instance in instances.values() if
            instance.name == instance_name]
```

giovannt0: Done

libcloudforensics/providers/aws/internal/ebs.py
Fryyyyy: I still don't think having assert statements in production code is a great idea.
Even if it's to satisfy the checker, it doesn't hurt us to have an actual if statement, that has no danger of being bypassed by someone's environment.

giovannt0: @Fryyyyy as discussed earlier, replaced other occurrences. PTAL

PR #234, opened 2020-08-02
Refactor Azure module

### LINKED ISSUE #224: Refactor Azure internal module
Similar to #222, in Azure

giovannt0:
Closes #224. 

Together with #225, this PR harmonizes the structure under which cloud provider functionality are implemented, following GCP's module organisation. `account.py` now only contains references to modules (compute, networking, monitoring, storage, etc). 

This breaks down the code a little more and separate functionality better.

Unit tests, end to end tests and CLI files are updated as well to use the new structure.

### DIFF
--- a/libcloudforensics/providers/azure/internal/compute.py
+++ b/libcloudforensics/providers/azure/internal/compute.py
@@ -432,7 +432,7 @@ class AZCompute:
             'hardwareProfile': {'vmSize': instance_type},
             'storageProfile': {
                 'imageReference': {
-                    'sku': '18.04-LTS',
+                    'sku': common.UBUNTU_1804_SKU,
                     'publisher': 'Canonical',
                     'version': 'latest',
                     'offer': 'UbuntuServer'}
--- a/libcloudforensics/providers/azure/internal/compute.py
+++ b/libcloudforensics/providers/azure/internal/compute.py
@@ -12,20 +12,28 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Azure Compute Resources."""
+"""Azure Compute functionality."""
 
+import base64
+import hashlib
 from time import sleep
-from typing import Optional, List, Dict, TYPE_CHECKING
+from typing import Optional, List, Dict, TYPE_CHECKING, Tuple, Any
 
 # Pylint complains about the import but the library imports just fine,
 # so we can ignore the warning.
 # pylint: disable=import-error
-from azure.mgmt.compute.v2020_05_01 import models
+import sshpubkeys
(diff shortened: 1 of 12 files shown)


libcloudforensics/providers/azure/internal/network.py
```diff
@@ -0,0 +1,180 @@
...
+    network_interface_name = '{0:s}-nic'.format(name)
+    ip_config_name = '{0:s}-ipconfig'.format(name)
+
+    # Check if the network interface already exists, and returns its ID if so.
+    try:
+      nic = self.network_client.network_interfaces.get(
+          self.az_account.default_resource_group_name, network_interface_name)
+      nic_id = nic.id  # type: str
+      return nic_id
+    except azure_exceptions.CloudError:
```
Fryyyyy, line 83: Is it not possible that CloudError would also encapsulate other error situations than just "the NIC already exists"?
If something else goes wrong, we just eat the exception and continue..

libcloudforensics/providers/azure/internal/compute.py
```diff
@@ -36,47 +44,508 @@
...
+    if not region:
+      region = self.az_account.default_region
+
+    creation_data = {
+        'location': region,
+        'properties': {
+            'hardwareProfile': {'vmSize': instance_type},
+            'storageProfile': {
+                'imageReference': {
+                    'sku': '18.04-LTS',
```
Fryyyyy, line 435: Is this something we want to put in the code rather than config?
And if so, do we want it buried deep within a data blob rather than a constant at the top?

libcloudforensics/providers/azure/internal/compute_base_resource.py
```diff
@@ -0,0 +1,74 @@
...
+      name (str): The resource's name.
+      region (str): The region in which the resource is located.
+      zones (List[str]): Optional. Availability zones within the region where
+          the resource is located.
+    """
+
+    self.az_account = az_account
+    # Format of resource_id: /subscriptions/{id}/resourceGroups/{
+    # resource_group_name}/providers/Microsoft.Compute/{resourceType}/{resource}
+    self.resource_group_name = resource_id.split('/')[4]
```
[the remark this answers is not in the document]
giovannt0: Added a regex check, and raise a ValueError if it's malformed

libcloudforensics/providers/azure/internal/network.py
giovannt0: Changed to raise the exception if it's not about the NIC not being found (which is the one we wanna ignore)

libcloudforensics/providers/azure/internal/compute.py
giovannt0: Moved to `common.py`

giovannt0:
Thanks @Fryyyyy, PTAL

PR #236, opened 2020-08-04
Auto-create network security group when VM is created in Azure

giovannt0:
Since the official Azure CLI tool creates a default network security group when creating a VM (e.g. through `az vm create --name=vm_name --resource-group=resource_group_name --image=UbuntuLTS --admin-username=azureuser --generate-ssh-keys`), add the same functionality to libcloudforensics, so that only SSH connection is allowed on the VM. Other outbound / inbound connexions are denied by default (see https://docs.microsoft.com/en-us/azure/virtual-network/security-overview#default-security-rules)


libcloudforensics/providers/azure/internal/network.py
```diff
@@ -87,7 +87,7 @@ def CreateNetworkInterface(self,
       # NIC doesn't exist, ignore the error and create it
 
     # pylint: disable=unbalanced-tuple-unpacking
-    public_ip, _, subnet = self._CreateNetworkInterfaceElements(
+    public_ip, _, subnet, nsg = self._CreateNetworkInterfaceElements(
```
rgayon, line 90: nit: maybe make the name more explicit, ie netsecuritygrp ?

libcloudforensics/providers/azure/internal/network.py
```diff
@@ -141,6 +143,7 @@ def _CreateNetworkInterfaceElements(
     public_ip_name = '{0:s}-public-ip'.format(name_prefix)
     vnet_name = '{0:s}-vnet'.format(name_prefix)
     subnet_name = '{0:s}-subnet'.format(name_prefix)
+    nsg_name = '{0:s}-nsg'.format(name_prefix)
```
rgayon, line 146: same for naming

libcloudforensics/providers/azure/internal/network.py
giovannt0: `NSG` is the official Azure abbreviation so if possible I'd rather stick with this, but I added a comment above that line (and the line for the next comment) to make it a little clearer. Let me know what you think!

rgayon: SGTM, thanks!

PR #244, opened 2020-08-13
adding a default zone for a GCP project

sa3eed3ed:
fixing the docstring example in gcp project class and adding a default zone, Global zone means nothing in gcP, zonal resources will need a zone


libcloudforensics/providers/gcp/internal/project.py
```diff
@@ -35,19 +35,19 @@ class GoogleCloudProject:
...
   Example use:
-    gcp = GoogleCloudProject("your_project_name", "us-east")
+    gcp = GoogleCloudProject("your_project_name", "us-east1-b")
     gcp.compute.ListInstances()
   """
 
   def __init__(self,
                project_id: str,
-               default_zone: Optional[str] = None) -> None:
+               default_zone: Optional[str] = 'us-central1-f') -> None:
```
giovannt0, line 44: If there's a default value for a parameter then the `Optional` type hint keyword is to be dropped: `str: 'us-central1-f`

PR #266, opened 2020-10-12
Add deleting instance functionality

### LINKED ISSUE #265: Consider adding deleteInstance/deleteDisks
This would be nice to have when implementing CI tests, and we want a way to cleanup stuff

rgayon:
Fixes #265 by adding a Delete() method to the gce Disk & Instance objects.


tools/cli.py
```diff
@@ -309,6 +310,13 @@ def Main() -> None:
                 ('--attach_disks', 'Comma separated list of disk names '
                                    'to attach.', None)
             ])
+  AddParser('gcp', gcp_subparsers, 'deleteinstance', 'Delete a GCE instance.',
+            args=[
+                ('instance_name', 'Name of the GCE instance to delete.', ''),
+                ('--delete_disks',
+                 'Whethere to delete all disks attached to the instance.',
```
sa3eed3ed, line 317: typo Whethere -> Whether
and I think 'Whether to force delete all disks attached to the instance.' since it's only about deleting the RO and the 'don't delete when instance is deleted' disks. To no give the false impression that keeping this False will not delete attached disks

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -854,6 +854,35 @@ def DetachDisk(self, disk: 'GoogleComputeDisk') -> None:
...
+        (ignores the 'readonly' bit).
+    """
+    disks_to_delete = []
+    if delete_disks:
+      disks_to_delete = [
+          disk['source'].split('/')[-1] for disk in self.GetValue('disks')]
+
+    gce_instance_client = self.GceApi().instances()
+    logger.info(
+        self.FormatLogMessage('Deleted Instance: {0:s}'.format(self.name)))
```
sa3eed3ed, line 871: maybe Deleting? since the deletion may fail and this will leave a confusing log line.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -854,6 +854,35 @@ def DetachDisk(self, disk: 'GoogleComputeDisk') -> None:
...
+    request = gce_instance_client.delete(
+        project=self.project_id, instance=self.name, zone=self.zone)
+    response = request.execute()
```
sa3eed3ed, line 874: add the delete instance in a try/except block to catch the resource not found error.
```suggestion
    try:
      request = gce_instance_client.delete(
          project=self.project_id, instance=self.name, zone=self.zone)
        response = request.execute()
    except HttpError as error:
      if error.resp.status == 404:
        log.warning(('Can not find resource {0:s}, the it might be already '
            'deleted. API call resulted in the following error: '
            '{1:s}').format(self.name, str(error)))
      else:
        log.error(
            'While deleting resource {0:s} the following error occurred: {1:s}'
            .format(self.name, str(error))) 
```

sa3eed3ed (changes requested):
little things

sa3eed3ed:
@rgayon can you please add unit-tests or open a new issue and add them in a subsequent PR?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -915,6 +944,17 @@ def Snapshot(self,
...
+  def Delete(self) -> None:
+    """Delete a Disk."""
+
+    gce_disk_client = self.GceApi().disks()
+    request = gce_disk_client.delete(
+        project=self.project_id, disk=self.name, zone=self.zone)
+    response = request.execute()
+    logger.info(
+        self.FormatLogMessage('Deleted Disk: {0:s}'.format(self.name)))
+    return response
```
sa3eed3ed, line 956: remove the return or change the return type above, linter is complaining

tools/cli.py
rgayon: agreed. Let me change this a bit to be more explicit

libcloudforensics/providers/gcp/internal/compute.py
rgayon: done

libcloudforensics/providers/gcp/internal/compute.py
rgayon: Done.
Did the same with the disk.Delete() method

libcloudforensics/providers/gcp/internal/compute.py
rgayon: done

rgayon:
> @rgayon can you please add unit-tests or open a new issue and add them in a subsequent PR?

So I actually wonder what I should be testing for, if I'm mock'ing gce_image_client.delete() & gce_disk_client.delete() as I'm not using their result whatsoever.
This is similar to how gce_image_client.delete() & gce_snapshot_client.delete() don't have unit tests either

sa3eed3ed:
> > @rgayon can you please add unit-tests or open a new issue and add them in a subsequent PR?
> 
> So I actually wonder what I should be testing for, if I'm mock'ing gce_image_client.delete() & gce_disk_client.delete() as I'm not using their result whatsoever.
> This is similar to how gce_image_client.delete() & gce_snapshot_client.delete() don't have unit tests either

yes thats true, I think what can be tested here is the delete_disks flag to make sure that disks_to_delete is what is expected to get (from the mock with disks.source fields ).

rgayon:
> > > @rgayon can you please add unit-tests or open a new issue and add them in a subsequent PR?
> > 
> > 
> > So I actually wonder what I should be testing for, if I'm mock'ing gce_image_client.delete() & gce_disk_client.delete() as I'm not using their result whatsoever.
> > This is similar to how gce_image_client.delete() & gce_snapshot_client.delete() don't have unit tests either
> 
> yes thats true, I think what can be tested here is the delete_disks flag to make sure that disks_to_delete is what is expected to get (from the mock with disks.source fields ).

Added a test to check that the disks().delete() method is called

rgayon:
> little things

ack

sa3eed3ed (approved):

hacktobeer:
sa3eed3ed did a solid review, lgtm from me.

hacktobeer (approved):

PR #271, opened 2020-11-23
Add device name to attachDisk calls

aarontp:
This is so that we can have a predictable device name in the instance (which is needed by Turbinia).


hacktobeer:
LGTM

Getting below build error but that's not due to this (small) code change.
"error: PyJWT 2.0.0a1 is installed but PyJWT[crypto]<2,>=1.0.0 is required by {'msal'}"

hacktobeer (approved):

tomchop:
Thanks for looking into the reason the tests are failing @hacktobeer. I'll merge these changes and do a dep bump

PR #275, opened 2021-01-05
Add instance network quarantine function

sa3eed3ed:
Add instance network quarantine function to GCP and CLI, with a list of exempted IPs.

### DIFF
--- a/libcloudforensics/providers/gcp/forensics.py
+++ b/libcloudforensics/providers/gcp/forensics.py
@@ -15,6 +15,7 @@
 """Forensics on GCP."""
 
 import base64
+import random
 from typing import TYPE_CHECKING, List, Tuple, Optional, Dict, Any
 
 from google.auth.exceptions import RefreshError, DefaultCredentialsError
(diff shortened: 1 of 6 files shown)


libcloudforensics/providers/gcp/forensics.py
```diff
@@ -227,3 +227,47 @@ def CreateDiskFromGCSImage(
       'md5Hash': md5_hash_hex
   }
   return result
+
+
+def InstanceNetworkQuarantine(project_id: str,
+                        instance_name: str,
+                        exempted_src_ips: Optional[List[str]] = None,
+                        enable_logging: bool = False) -> None:
```
tomchop, line 235: Please fix indentation here.

libcloudforensics/providers/gcp/forensics.py
```diff
@@ -227,3 +227,47 @@ def CreateDiskFromGCSImage(
...
+  """Put a Google Cloud instance in network quarantine.
+
```
tomchop, line 237: Can you be a bit more verbose about what network quarantine implies here? e.g. moving to another VPC network, applying firewall rules, etc.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -697,6 +697,97 @@ def ImportImageFromStorage(self,
...
+    """
+
+    logger.info( 'Inserting firewall rule {0:s}, '
+            'targeting tags: {1!s}.'.format(body['name'], body['targetTags'] ))
+    firewall_client = self.GceApi().firewalls()
+    request = firewall_client.insert(project=self.project_id, body=body)
+    response = request.execute()
+    self.BlockOperation(response)
+
+  def AddDenyAllFirewallRules(self,
```
tomchop, line 715: I think this function would be better placed in `forensics.py`, as it contains mostly "DFIR logic"

libcloudforensics/providers/gcp/forensics.py
```diff
@@ -227,3 +227,47 @@ def CreateDiskFromGCSImage(
...
+  instance = project.compute.GetInstance(instance_name)
+  get_operation = instance.GetOperation()
+  network_interfaces = get_operation['networkInterfaces']
+  target_tags = []
+  for interface in network_interfaces:
+    network_name = interface["network"]
+    deny_ingress_tag = common.GenerateUniqueInstanceName(
+        'deny-ingress-tag-', common.COMPUTE_NAME_LIMIT)
+    deny_egress_tag = common.GenerateUniqueInstanceName(
+        'deny-egress-tag-', common.COMPUTE_NAME_LIMIT)
```
tomchop, line 258: Why do we need tags for here? Why Are we calling `GenerateUniqueInstanceName` if we're looking to generate a tag name?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -697,6 +697,97 @@ def ImportImageFromStorage(self,
...
+            Default is False.
+
+      Raises:
+        InvalidNameError: If Tag names are invalid.
+    """
+
+    logger.info('Creating deny-all (ingress/egress) '
+            'firewall rules in {0:s} network.'.format(network))
+
+    if not common.REGEX_DISK_NAME.match(deny_ingress_tag):
```
tomchop, line 739: We should probably come up with a better name than `REGEX_DISK_NAME` if we're going to use it on tags too.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -897,6 +988,47 @@ def Delete(self, delete_disks: bool = False) -> None:
             self.FormatLogMessage(
                 'Could not find disk: {0:s}, skipping'.format(disk_name)))
 
+  def SetTags(self, new_tags: List[str]) -> None:
+    """Sets network tags for the compute instance.
+
+    Args:
+        new_tags (List[str]): A list of tags. Each tag must be 1-63
```
tomchop, line 995: Please fix indentation.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -897,6 +988,47 @@ def Delete(self, delete_disks: bool = False) -> None:
...
+        self.FormatLogMessage(', adding tags {0!s} to instance '
+            '{1:s}.'.format(new_tags, self.name)))
+    for tag in new_tags:
+      if not common.REGEX_DISK_NAME.match(tag):
+        raise errors.InvalidNameError(
+            'Network Tag {0:s} does not comply with {1:s}.'.format(
+                tag, common.REGEX_DISK_NAME.pattern), __name__)
+
+    get_operation = self.GetOperation()
+    tags_object = get_operation['tags']
```
tomchop, line 1013: This is a dict rather than an object?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -897,6 +988,47 @@ def Delete(self, delete_disks: bool = False) -> None:
             self.FormatLogMessage(
                 'Could not find disk: {0:s}, skipping'.format(disk_name)))
 
+  def SetTags(self, new_tags: List[str]) -> None:
+    """Sets network tags for the compute instance.
```
tomchop, line 992: Is this only for *network* tags? It seems to be generic for any kind of tags.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -897,6 +988,47 @@ def Delete(self, delete_disks: bool = False) -> None:
...
+
+    Args:
+        new_tags (List[str]): A list of tags. Each tag must be 1-63
+            characters long, and comply with RFC1035.
+
+    Raises:
+      InvalidNameError: If the name of the snapshot does not
```
tomchop, line 999: Do you mean `tags` instead of `snapshot`?

tools/gcp_cli.py
```diff
@@ -285,3 +285,22 @@ def DeleteObject(args: 'argparse.Namespace') -> None:
...
+
+    Args:
+      args (argparse.Namespace): Arguments from ArgumentParser.
+  """
+  exempted_ips = []
+  if args.exempted_src_ips:
+    exempted_ips = args.exempted_src_ips.split(',')
+    # Check if exempted_src_ips argument exists and if there
+    # are any empty entries.
+    if not (exempted_ips and all(elements for elements in exempted_ips)):
```
tomchop, line 301: I *think* you can do `all(exempted_ips)`
```suggestion
    if not (exempted_ips and all(exempted_ips)):
```

tomchop (changes requested):
This is really cool! Do the firewall rules also isolate the instance from other instances in the same network? If not, you could create a "quarantine" VPC network and just move the instance there with appropriate firewall rules.

tools/gcp_cli.py
sa3eed3ed: oh, that is indeed right

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: yes it is tags, I was so lazy and copied part of the error description from another docstring

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: tags are only used to configure firewall rules and network routes, in some reference they are referred to as network tags but not in every case. I removed network from the description and added a new description line to clarify that these are the ones used for firewall/routes management.
https://cloud.google.com/vpc/docs/add-remove-network-tags

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: I changed it to dict. I meant API object, but dict is consistent with the rest of the code.

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: added a new COMPUTE_RFC1035_REGEX constant to the common.py module and referenced REGEX_DISK_NAME. Left  REGEX_DISK_NAME to not change it everywhere.

libcloudforensics/providers/gcp/forensics.py
sa3eed3ed: just to add a time stamp to the prefix. I used this to guarantee that the tag name will not collide with any other tag name, specially if there are multiple network interfaces, which means we have to add more tags to the same instance. I also used the same tag name as a firewall name later-on to guarantee it doesn't collide with other FW rule names

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: Done

libcloudforensics/providers/gcp/forensics.py
sa3eed3ed: Done

sa3eed3ed:
> This is really cool! Do the firewall rules also isolate the instance from other instances in the same network? If not, you could create a "quarantine" VPC network and just move the instance there with appropriate firewall rules.

Yes the firewall rules isolate the instance from other instances on the same network and on multiple networks, if the instance has multiple interfaces attached. I added the exempted IP list functionality to allow the analyst connect to the machine if needed, since connecting from the same project/network won't be allowed anymore.

tomchop (approved):

hacktobeer (approved):
Exciting new feature, thanks Said!

PR #288, opened 2021-03-31
Add sleep() in Azure e2e to allow for deleteOperation propagation

giovannt0:
- Fix e2e test, where sometimes a resource's deletion is propagated too slow and tests depending on similar resources fail.
- Fix Azure unit tests versioning
- Fix GitHub action for py3.6 where the crypto package needs to be installed first as otherwise it fails in the global setup.

### DIFF
--- a/tests/providers/azure/e2e.py
+++ b/tests/providers/azure/e2e.py
@@ -143,11 +143,11 @@ class EndToEndTest(unittest.TestCase):
     # permit same-name disks in different regions.
     operation = self.az.compute.compute_client.disks.begin_delete(
         disk_copy.resource_group_name, disk_copy.name)
-    operation.wait()
     # Delete operation takes some time to propagate within Azure, which in
-    # some cases causes the next test to fail. Therefore we wait a minute
-    # before moving on to the next test.
-    sleep(60)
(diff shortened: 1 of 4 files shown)


tests/providers/azure/e2e.py
```diff
@@ -144,6 +144,10 @@ def testDiskCopyToOtherZone(self):
     operation = self.az.compute.compute_client.disks.begin_delete(
         disk_copy.resource_group_name, disk_copy.name)
     operation.wait()
+    # Delete operation takes some time to propagate within Azure, which in
+    # some cases causes the next test to fail. Therefore we wait a minute
+    # before moving on to the next test.
+    sleep(60)
```
hacktobeer, line 150: If we can't make this more elegant (actually a loop that checks if it's removed or not) I am good with this.

hacktobeer (approved):

tests/providers/azure/e2e.py
giovannt0: Elegant-ified :)

hacktobeer: nice and thanks!

rgayon (approved):

Issue #164, opened 2020-06-23
Create e2e tests for the CLI tools.


hacktobeer:

hacktobeer:
Look into using GCP cloud build for e2e tests.

PR #313, opened 2021-06-03
Add e2e tests for GCP CLI tool

giovannt0:
As part of addressing #164, starting with GCP. The tests mostly follow what we already have in place for e2e tests, the only difference being that operations are delegated to the CLI tool in place of the `forensics` module.


Fryyyyy:
Why are these called *bash ?

giovannt0:
@Fryyyyy because I'm dumb and forgot to change the files name (I was initially gonna do some bash stuff hence the name, but changed my mind). Fixed!

tests/providers/gcp/e2e_cli.py
```diff
@@ -0,0 +1,275 @@
...
+    logger.info('Deleting analysis instance: {0:s}.'.format(
+        analysis_vm.name))
+    gce_instance_client = gcp_client.GceApi().instances()
+    request = gce_instance_client.delete(
+        project=project.project_id, zone=zone, instance=analysis_vm.name)
+    response = request.execute()
+    try:
+      gcp_client.BlockOperation(response)
+    except HttpError:
+      # BlockOperation triggers a while(True) loop that checks on the
```
Fryyyyy, line 237: Should we have a bug/issue open that tracks this and changes it to a tighter While loop (on a condition that changes when the operation has finished?)

Fryyyyy (approved):

tests/providers/gcp/e2e_cli.py
giovannt0: Agreed, I'll open one, thanks!

PR #314, opened 2021-06-04
Add e2e tests for AWS CLI tool

giovannt0:
Following #313, 2nd out of 3 parts for fixing #164


tests/providers/aws/aws_cli.py
```diff
@@ -0,0 +1,167 @@
...
+    try:
+      output = subprocess.check_output(
+          cmd.split(), stderr=subprocess.STDOUT, shell=False)
+      logger.info(output)
+      try:
+        return self.aws.ec2.GetInstancesByName(vm_name)[0]
+      except ResourceNotFoundError as error:
+        raise ResourceNotFoundError(
+            'Failed finding created VM in account profile {0:s}'.format(
+                self.aws.aws_profile), __name__) from error
```
tomchop, line 75: Can we move this to outside of the wrapping `try` block? It seems unlikely that this part would ever raise a `CalledProcessError` exception.

Also - should this part go in the actual test?

tests/providers/aws/aws_cli.py
```diff
@@ -0,0 +1,167 @@
...
+  """AWSCLI calls the libcloudforensics CLI tool for AWS operations."""
+  def __init__(self, aws: AWSAccount) -> None:
+    """Initialize the CLI class.
+
+    Attributes:
+      aws (AWSAccount): The AWS account to work with.
+    """
+    self.aws = aws
+
+  def StartAnalysisVm(
```
tomchop, line 40: Just for clarity, can we rename this to "RunStartAnalysisVm" command? And adjust docstrings to emphasize that this is just a wrapper around the CLI?

tests/providers/aws/aws_cli.py
```diff
@@ -0,0 +1,167 @@
...
+      volume_copy_name = volume_copy_name[:volume_copy_name.rindex('copy') + 4]
+      logger.info(output)
+      logger.info("Volume successfully copied to {0:s}".format(
+          volume_copy_name))
+      try:
+        return self.aws.ebs.GetVolumesByName(volume_copy_name)[0]
+      except ResourceNotFoundError as error:
+        raise ResourceNotFoundError(
+            'Failed finding copied volume in account profile {0:s}'.format(
+                self.aws.aws_profile), __name__) from error
```
tomchop, line 132: Same thing here, and in general. I would limit these helper functions (e.g. `CreateVolumeCopy`) to just creating the command you need, and then do all the checking as part of the test itself.

tests/providers/aws/aws_cli.py
```diff
@@ -0,0 +1,167 @@
...
+from libcloudforensics.errors import ResourceNotFoundError
+from libcloudforensics.providers.aws.internal import ebs
+from libcloudforensics.providers.aws.internal import ec2
+from libcloudforensics.providers.aws.internal.account import AWSAccount
+
+logging_utils.SetUpLogger(__name__)
+logger = logging_utils.GetLogger(__name__)
+
+
+class AWSCLI:
```
tomchop, line 30: Can we rename this to AWSCLIHelper, and indicate in the docstring that it's just a set of helper functions to run the CLI invocations?

tomchop (changes requested):

tests/providers/aws/aws_cli.py
giovannt0: Moved the block.

Re: `Also - should this part go in the actual test?` Great question. I tried to make the actual e2e CLI file as close as possible as the Python one, hence why this code is here. The other option is to return whatever we need from the output of the CLI call (mostly, instance + disk names as strings) and then grab the instance / disk as a Python object in the test directly. I don't have a strong preference for either option, so up to you.

tests/providers/aws/aws_cli.py
giovannt0: ACK, will update

tests/providers/aws/e2e_cli.py
```diff
@@ -0,0 +1,332 @@
...
+      if not output:
+        raise ResourceCreationError(
+            'Could not find volume copy result', __name__)
+      volume_copy_name = output.decode('utf-8').split(' ')[-1]
+      volume_copy_name = volume_copy_name[:volume_copy_name.rindex('copy') + 4]
+      logger.info(output)
+      logger.info("Volume successfully copied to {0:s}".format(
+          volume_copy_name))
+    except subprocess.CalledProcessError as error:
+      raise ResourceCreationError('Failed copying volume', __name__) from error
```
tomchop, line 277: You want this to be as close to the `subprocess` call as possible (no use to have the decode / logging operations inside this try block)

tests/providers/aws/e2e_cli.py
```diff
@@ -0,0 +1,332 @@
...
+  @typing.no_type_check
+  @IgnoreWarnings
+  def testListImages(self):
+    """End to end test on AWS.
+
+    Test listing AMI images with a filter.
+    """
+
+    qfilter = 'ubuntu*'
+    images = self._RunListImages(self.zone, qfilter)
```
tomchop, line 157: Looks like _RunListImages is only called once throughout the whole file. Why make it a separate function? I think it would be more clear to have everything in the test function.

tests/providers/aws/e2e_cli.py
```diff
@@ -0,0 +1,332 @@
...
+    if not (self.volume_to_copy and self.dst_zone):
+      return
+
+    volume_copy = self._RunCreateVolumeCopy(
+        self.zone, dst_zone=self.dst_zone, volume_id=self.volume_to_copy)
+    # The volume should be created in AWS
+    aws_account = account.AWSAccount(self.dst_zone)
+    aws_volume = aws_account.ResourceApi(EC2_SERVICE).Volume(
+        volume_copy.volume_id)
+    self.assertEqual(aws_volume.volume_id, volume_copy.volume_id)
```
tomchop, line 145: Could we also assert that the zones of `aws_volume` and `volume_copy` are indeed different? Same comment is valid for other tests - besides the code not erroring out, are there any ways we can confirm that the operation is working as intended?

tests/providers/aws/e2e_cli.py
```diff
@@ -0,0 +1,332 @@
...
+
+    volume_copy = self._RunCreateVolumeCopy(
+        self.zone,
+        dst_zone=self.dst_zone,
+        volume_id=self.encrypted_volume_to_copy)
+    # The volume should be created in AWS
+    aws_account = account.AWSAccount(self.dst_zone)
+    aws_volume = aws_account.ResourceApi(EC2_SERVICE).Volume(
+        volume_copy.volume_id)
+    self.assertEqual(aws_volume.volume_id, volume_copy.volume_id)
```
tomchop, line 200: Any way we can check that both volumes are encrypted and in different zones?

tests/providers/aws/e2e_cli.py
```diff
@@ -0,0 +1,332 @@
...
+    """
+
+    if not self.encrypted_volume_to_copy:
+      return
+
+    volume_copy = self._RunCreateVolumeCopy(
+        self.zone, volume_id=self.encrypted_volume_to_copy)
+    # The volume should be created in AWS
+    aws_volume = self.aws.ResourceApi(EC2_SERVICE).Volume(volume_copy.volume_id)
+    self.assertEqual(aws_volume.volume_id, volume_copy.volume_id)
```
tomchop, line 177: Any way we can check that both volumes are encrypted?

tomchop (changes requested):

tomchop (approved):

Issue #316, opened 2021-06-10
AWS startvm fails if there is no default VPC


ramo-j:
If an account has no default VPC, there is currently no way to specify one when creating a VM. This results in an error:

```
$ cloudforensics aws ap-southeast-2a startvm ramojlcftest 
[2021-06-10 05:37:08,940] [tools.aws_cli       ] INFO     Starting analysis VM...
[2021-06-10 05:37:08,946] [libcloudforensics.providers.aws.forensics] INFO     No AMI provided, fetching one for Ubuntu 18.04
[2021-06-10 05:37:09,588] [libcloudforensics.providers.aws.forensics] INFO     Starting analysis VM ramojlcftest
[2021-06-10 05:37:09,990] [libcloudforensics.providers.aws.internal.ec2] ERROR    Could not create instance ramojlcftest: An error occurred (VPCIdNotSpecified) when calling the RunInstances operation: No default VPC for this user. GroupName is only supported for EC2-Classic and default VPC.
Traceback (most recent call last):
  File "/home/ramoj_google_com/.local/lib/python3.7/site-packages/libcloudforensics-20210531-py3.7.egg/libcloudforensics/providers/aws/internal/ec2.py", line 413, in GetOrCreateAnalysisVm
    instance = client.run_instances(**vm_args)
  File "/home/ramoj_google_com/.local/lib/python3.7/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/ramoj_google_com/.local/lib/python3.7/site-packages/botocore/client.py", line 676, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (VPCIdNotSpecified) when calling the RunInstances operation: No default VPC for this user. GroupName is only supported for EC2-Classic and default VPC.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/ramoj_google_com/.local/bin/cloudforensics", line 11, in <module>
    load_entry_point('libcloudforensics==20210531', 'console_scripts', 'cloudforensics')()
  File "/home/ramoj_google_com/.local/lib/python3.7/site-packages/libcloudforensics-20210531-py3.7.egg/tools/cli.py", line 422, in Main
    parsed_args.func(parsed_args)
[...]

Issue #315, opened 2021-06-08
Make setting UserData in GetOrCreateAnalyssiVM() configurable


hiddenillusion:
`GetOrCreateAnalysisVM()` in the AWS ec2.py lib is currently hard coded to install the contents of `startup_script`.

It would be more flexible if this was determined via an argument since we may not want to install any supplemental packages (right now you can only remove the contents of `scripts/startup.sh` to satisfy this)

ramo-j:
There is currently an alternate method to specify a userdata script that is provided, via an env var. From the code comments on `utils.ReadStartupScript`:

```
Users can either write their own script to install custom packages,
or use one of the provided ones. To use your own script, export a
STARTUP_SCRIPT environment variable with the absolute path to it:
"user@terminal:~$ export STARTUP_SCRIPT='absolute/path/script.sh'"
```

That said, I agree that being able to pass in an argument makes more sense, but the above might suit you in the meantime.

PR #317, opened 2021-06-11
Fixes for #315, #316

ramo-j:
#315 - Allow specification of a subnet id to launch the instance in
#316 - An easier method for a user to provide a userdata script, and future support for additional scripts


libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -378,15 +380,6 @@ def GetOrCreateAnalysisVm(
       return instances[0], created
 
     instance_type = common.GetInstanceTypeByCPU(cpu_cores)
-    startup_script = utils.ReadStartupScript()
-    if packages:
-      startup_script = startup_script.replace('${packages[@]}', ' '.join(
-          packages))
-
-    # Install ec2-instance-connect to allow SSH connections from the browser.
-    startup_script = startup_script.replace(
```
giovannt0, line 387: Mmmh what happens here if `userdata` below is not provided? IIRC if you don't install `ec2-instance-connect` then you won't be able to SSH to the machine from the browser. Has this behaviour change? It looks like now the default behaviour is to not provide a startup script anymore. Is this assumption correct?

tools/aws_cli.py
```diff
@@ -155,6 +156,17 @@ def StartAnalysisVm(args: 'argparse.Namespace') -> None:
         'Created key pair {0:s} in AWS. Your private key is saved in: '
         '{1:s}'.format(key_name, path))
 
+  # Load up the startup script
+  startup_script = None
```
giovannt0, line 160: Since `startup_script` is assigned in both clauses of the `if` block, this can be removed

tools/aws_cli.py
```diff
@@ -155,6 +156,17 @@ def StartAnalysisVm(args: 'argparse.Namespace') -> None:
...
+  if args.launch_script:
+    startup_script = utils.ReadStartupScript(args.launch_script)
+  else:
+    startup_script = utils.ReadStartupScript(utils.FORENSICS_STARTUP_SCRIPT)
+    # Install ec2-instance-connect to allow SSH connections from the browser.
+    startup_script = startup_script.replace(
```
giovannt0, line 166: I see. This is for the CLI tool, though. What happens to the forensics package w.r.t. `ec2-instance-connect`? Shouldn't what's below also be surfaced in that package?

giovannt0 (changes requested):

libcloudforensics/providers/aws/forensics.py
```diff
@@ -285,3 +292,5 @@ def StartAnalysisVm(
         aws_account.ebs.GetVolumeById(volume_id), device_name)
   logger.info('VM ready.')
   return analysis_vm, created
+
+# pylint: enable=too-many-arguments
```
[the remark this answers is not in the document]
ramo-j: It looks like it needs to be at the end of the function definition, or the linter still throws the complaint about too many args.

tools/aws_cli.py
ramo-j: The end goal here is to be able to support different launch scripts. Perhaps the best answer is to have a forensics package script for each aws, gcp and azure, and we can include `ec2-instance-connect` specifically in the aws one.

libcloudforensics/providers/aws/forensics.py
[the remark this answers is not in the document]
giovannt0: ACK, thanks

tools/aws_cli.py
giovannt0: Sure, but now if you call `aws.forensics.StartAnalysisVm`, in its current state `userData` is `None` and hence `ec2-instance-connect` won't be installed, which is a change in behaviour compared to what CFU did before. This is problematic because `ssh_key_name` is also `None` by default, hence the resulting VM will not be SSHable into if neither a key or the package is provided. What you added in `aws_cli.py` should also be added to `forensics.py` to keep the behaviour we had before but provide the choice of specifying a different startup script.

libcloudforensics/scripts/utils.py
```diff
@@ -38,14 +38,13 @@ def ReadStartupScript(filename: str = '') -> str:
   """
 
   try:
-    script_path = None
```
giovannt0: This one needs to stay, because if `filename` is not `''` then the second `if` block is gonna throw an error as `script_path` isn't defined

tools/aws_cli.py
ramo-j: Righto, readjusted to have multiple startup scripts for each provider, and `aws.forensics.StartAnalysisVm` has a default value for the userdata file being the existing AWS one. This should achieve no unexpected changes to anyone directly invoking that method, and allow custom scripts to be provided.

libcloudforensics/providers/aws/internal/ec2.py
ramo-j: See other pr convo.

libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -361,6 +360,9 @@ def GetOrCreateAnalysisVm(
       tags (Dict[str, str]): Optional. A dictionary of tags to add to the
           instance, for example {'TicketID': 'xxx'}. An entry for the instance
           name is added by default.
+      subnet_id (str): Optional. Subnet to launch the instance in
+      userdata (str): Optional. String passed to the instance as a userdata
```
giovannt0, line 364: Maybe change this to 
```
    userdata_file (str): Optional. Filename to be read in as the userdata launch script.
```
So it's consistent with `aws.forensics`? + `.` at the end of the sentences.

libcloudforensics/providers/aws/forensics.py
```diff
@@ -268,6 +275,8 @@ def StartAnalysisVm(
     ami = ami_list[0]['ImageId']
   assert ami  # Mypy: assert that ami is not None
 
+  userdata = utils.ReadStartupScript(userdata_file)
```
giovannt0, line 278: After you've changed `userdata_file: Optional[str] = utils.FORENSICS_STARTUP_SCRIPT_AWS` to `userdata_file: Optional[str] = None`, you can add above this line:
```
if not userdata_file:
  userdata_file = utils.FORENSICS_STARTUP_SCRIPT_AWS
```

The reason is that currently in`aws_cli.py` you are passing `args.launch_script` to the function directly, which defaults to `None` and will thus override `userdata_file: Optional[str] = utils.FORENSICS_STARTUP_SCRIPT_AWS`. This will cause `utils.ReadStartupScript()` to be called with `None` which in turn will cause the wrong script to be loaded (`STARTUP_SCRIPT` instead of `FORENSICS_STARTUP_SCRIPT_AWS`)

libcloudforensics/providers/aws/forensics.py
```diff
@@ -239,8 +243,11 @@ def StartAnalysisVm(
         will not be accessible. It is therefore recommended to fill in this
         parameter.
     tags (Dict[str, str]): Optional. A dictionary of tags to add to the
-          instance, for example {'TicketID': 'xxx'}. An entry for the instance
-          name is added by default.
+        instance, for example {'TicketID': 'xxx'}. An entry for the instance
+        name is added by default.
+    subnet_id (str): Optional. The subnet to launch the instance in
```
giovannt0, line 248: Super-nit for this and elsewhere in the changes you made: add a `.` at the end of the sentences. This helps making our auto-generated documentation on ReadTheDocs more readable :).

libcloudforensics/scripts/utils.py
```diff
@@ -15,18 +15,24 @@
...
   Users can either write their own script to install custom packages,
-  or use the provided one. To use your own script, export a STARTUP_SCRIPT
-  environment variable with the absolute path to it:
+  or use one of the provided ones. To use your own script, export a
+  STARTUP_SCRIPT environment variable with the absolute path to it:
   "user@terminal:~$ export STARTUP_SCRIPT='absolute/path/script.sh'"
 
+  Args:
+    filename (str): the name of the script in the scripts directory to read
+      Defaults to 'forensics_packages_startup.sh' if none specified
```
giovannt0, line 35: \+ `.`

giovannt0 (changes requested):
Thanks Ramoj, I have a couple more nits / fixes but otherwise we're good to go. I'll approve this and merge it once the changes are in.
Thanks a lot!

libcloudforensics/providers/aws/internal/ec2.py
ramo-j: It made more sense to me to read the userdata file in `forensics.py` and have the userdata string passed in to the vm creation method - in the context of a future feature addition which also will launch an instance, that isn't a forensics analysis vm.

libcloudforensics/providers/aws/forensics.py
ramo-j: Done. I tested that out, but only that it got _a_ script /facepalm

libcloudforensics/providers/aws/forensics.py
ramo-j: I think I got em all!

libcloudforensics/providers/aws/internal/ec2.py
giovannt0: Makes sense, thanks!

giovannt0 (approved):
LGTM, thanks!

PR #328, opened 2021-07-02
Fetch `project_id` from the default() environment

### LINKED ISSUE #248: GCP CLI should attempt to grab "project" from other locations
Looks like it requires `project` in order to properly become a func in `AddParser()` in `cli.py`, but the documentation doesn't easily show that this is a required field.

```
ArgumentParser(prog='cloudforensics gcp project listlogs', usage=None, description=None, formatter_class=<class 'argparse.HelpFormatter'>, conflict_handler='error', add_help=True)
```
otherwise the Namespace returned is just `Namespace(project='listlogs'`), which fails the `hasattr` check.

If I already have a project set (`gcloud config set project Project123`) then it would be helpful if it was able to pull said project value (or from the credentials file if we're able to authenticate that way at some point).

giovannt0:
Closes #248.


tools/gcp_cli.py
```diff
@@ -42,8 +43,14 @@ def ListInstances(args: 'argparse.Namespace') -> None:
 
   Args:
     args (argparse.Namespace): Arguments from ArgumentParser.
+
+  Raises:
+    AttributeError: If no project_id was provided and none was inferred
```
tomchop, line 48:
```suggestion
    AttributeError: If no project_id was provided and none could be inferred
```

tomchop (approved):
LGTM!

PR #331, opened 2021-07-02
Add option to force delete instances that have deletion protection in place in GCP and AWS

### LINKED ISSUE #283: Feature Request: Check for "Deletion Protection Enabled" and disable
Sometimes we aren't able to delete instances because Deletion Protection is enabled. This has to be manually turned off. If when deleting an instance we were able to check iteratively for a list of instances where it was enabled and then disable, it would make mass deletion easier.

Something like you pass the project ID and it checks the compute instance API, lists all instances then iteratively passes them back to check.

giovannt0:
Fixes #283.
Linter issues are addressed in https://github.com/google/cloud-forensics-utils/pull/329


libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -145,6 +145,43 @@ def AttachVolume(self,
 
     volume.device_name = device_name
 
+  def Delete(self, force_delete: bool = False) -> None:
+    """Delete an instance.
+
+    Args:
+      force_delete (bool): Optional. True if the instance should be deleted
+          despite DisableApiTermination being set to True on the instance.
```
tomchop, line 153: nit: isn't this lowercase `disableApiTermination` like on L167?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -872,19 +877,42 @@ def DetachDisk(self, disk: 'GoogleComputeDisk') -> None:
...
     """
     disks_to_delete = []
     if delete_disks:
       disks_to_delete = [
           disk['source'].split('/')[-1] for disk in self.GetValue('disks')]
 
     gce_instance_client = self.GceApi().instances()
+
+    if force_delete:
+      if self.deletion_protection:
```
tomchop, line 898:
```suggestion
    if force_delete and self.deletion_protection:
```

tomchop: Should we output a warning if `force_delete` is `False` and `self.deletion_protection` is `True` so the user knows to add the parameter?

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -872,19 +877,42 @@ def DetachDisk(self, disk: 'GoogleComputeDisk') -> None:
...
+              project=self.project_id,
+              zone=self.zone,
+              resource=self.name,
+              deletionProtection=False)
+          response = request.execute()
+        except HttpError as exception:
+          logger.error('Unable to toggle deleteProtection on instance {0:s}: '
+                       '{1:s}'.format(self.name, str(exception)))
+          raise errors.ResourceDeletionError(
+              'Could not delete instance {0:s}: {1!s}'.format(
```
tomchop, line 912: Given that we're handling an error related to `setDeletionProtection` and not the deletion itself, can we change this error message to reflect that the error happened while carrying out that operation?

tomchop (changes requested):

libcloudforensics/providers/gcp/internal/compute.py
giovannt0: Yep, added

tomchop (approved):

PR #335, opened 2021-07-09
Query GCP logs with multiple projects

### LINKED ISSUE #245: GCP log query should handle searching more than one resource
It would be useful if we were able to execute a single query against all available resources (projects, organizations etc.).

`GoogleCloudLog` currently accepts a single project ID and uses that for where to perform its search. This is probably fine for most use cases, but fails to address:

- A user wanting to perform a search across all available resources

- A user wanting to search across some combination of 1+ resource (e.g. 10 out of their 50 projects)

Based on the API docs, it looks like the [`resourceNames`](https://github.com/google/cloud-forensics-utils/blob/master/libcloudforensics/providers/gcp/internal/log.py#L102) field has the ability to accept a list of 1+ resource to query. 

If that's not the case, and while not ideal, it may be useful to at least have the ability to loop through X resources and issue said query on behalf of the user.

This improvement would help relieve some of the burden users face when using other options where they're limited to searching a single resource (`gcloud`, `Logs Viewer` etc.)... which usually results in something along the lines of looping through each project and executing said query (not ideal if you have hundreds of projects).

giovannt0:
Closes #245 

Add support for multiple project IDs for log listing and querying.

Example usage:

```
cloudforensics gcp --project=project1,project2 listlogs
cloudforensics gcp --project=project1,project2 querylogs
cloudforensics gcp --project=project1,project2 querylogs --filter="filter_expression_project_1,filter_expression_project_2"
```


libcloudforensics/providers/gcp/internal/log.py
```diff
@@ -26,7 +26,7 @@ class GoogleCloudLog:
   """Class representing a Google Cloud Logs interface.
 
   Attributes:
-    project_id: Google Cloud project ID.
+    project_ids: Comma-separated list of Google Cloud project IDs.
```
Fryyyyy, line 29: I think it would be nicer to have this as a Python list, since these really are libraries that we're providing and the CLI is just a reference interface...the CLI function can do the `split(',')`

Fryyyyy, line 29: Nit: update comment

Fryyyyy (approved):

PR #342, opened 2021-07-14
Do not cache the result of CreateService

tomchop:
Calls to `GoogleCloudFunction` would otherwise retain stale credentials if the object was reused to call `ExecuteFunction` over long periods of time

### DIFF
--- a/libcloudforensics/providers/gcp/internal/build.py
+++ b/libcloudforensics/providers/gcp/internal/build.py
@@ -28,9 +28,6 @@ class GoogleCloudBuild:
 
   Dictionary objects content can be found in
   https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.builds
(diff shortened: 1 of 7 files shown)


giovannt0 (commented):
If we're not gonna cache it, shouldn't we remove the `self.xxxx` attributes all-together and just directly return the result of the call to the service builder?

tomchop:
One could argue that it makes it easier to debug to have an intermediary variable, but I agree that we can remove it altogether.

giovannt0:
Yeah, I don't have a strong preference here, so I'll approve this and let you decide

giovannt0 (approved):

giovannt0:
(Pending build success)

tomchop:
Force merging this as we want these changes in sooner rather than later.

Issue #340, opened 2021-07-14
Implement quarantinevm for AWS


ramo-j:
`quarantinevm` exists for GCP - implement the same functionality for AWS.

PR #345, opened 2021-07-20
Implement quarantinevm for AWS - issues/340

ramo-j:
As per https://github.com/google/cloud-forensics-utils/issues/340

Adding the ability to network isolate an instance in AWS. This required the addition of vpc info to the EC2 object.

### DIFF
--- a/libcloudforensics/providers/aws/forensics.py
+++ b/libcloudforensics/providers/aws/forensics.py
@@ -447,7 +447,8 @@ def CopyEBSSnapshotToS3(
     logger.info(
       'Image copy timeout. The process may be ongoing, or might have failed.')
 
-def InstanceNetworkQuarantine(zone: str,
+def InstanceNetworkQuarantine(
+    zone: str,
     instance_id: str,
     exempted_src_subnets: Optional[List[str]] = None
     ) -> None:
@@ -459,13 +460,18 @@ def InstanceNetworkQuarantine(zone: str,
   Args:
     instance_id (str): : The id (i-xxxxxx) of the virtual machine.
     exempted_src_subnets (List[str]): List of subnets that will be permitted
+
+  Raises:
+    ResourceNotFoundError: If the instance cannot be found.
+    ResourceCreationError: If the security group could not be created.
+    AddressValueError: If a provided subnet is invalid.
   """
   # Add /32 to any specified subnets that don't have a mask
   # We're not checking the subnet is well formed, CreateIsolationSecurityGroup
   # will take care of that
   if exempted_src_subnets:
-    exempted_src_subnets[:] = [subnet if subnet.__contains__('/') else
-      subnet + '/32' for subnet in exempted_src_subnets]
+    exempted_src_subnets[:] = [subnet if '/' in subnet else subnet + '/32' 
+      for subnet in exempted_src_subnets]
 
   try:
     aws_account = account.AWSAccount(zone)
@@ -473,9 +479,7 @@ def InstanceNetworkQuarantine(zone: str,
     sg_id = \
       aws_account.ec2.CreateIsolationSecurityGroup(vpc, exempted_src_subnets)
     aws_account.ec2.SetInstanceSecurityGroup(instance_id, sg_id)
-  except errors.ResourceNotFoundError:
-    logger.error('Cannot qurantine non-existent instance')
-  except errors.ResourceCreationError:
-    logger.error('Failed to create security group')
-  except ipaddress.AddressValueError:
-    logger.error('Invalid subnet provided')
+  except errors.ResourceNotFoundError as exception:
+    raise errors.ResourceNotFoundError(
+      'Cannot qurantine non-existent instance {0:s}: {1!s}'.format(instance_id,
+        exception), __name__) from exception
--- a/libcloudforensics/providers/aws/forensics.py
+++ b/libcloudforensics/providers/aws/forensics.py
@@ -15,6 +15,7 @@
 """Forensics on AWS."""
 from typing import TYPE_CHECKING, Tuple, List, Optional, Dict
 
+import ipaddress
 import random
 from time import sleep
 from libcloudforensics.providers.aws.internal.common import ALINUX2_BASE_FILTER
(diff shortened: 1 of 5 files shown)


tools/cli.py
```diff
@@ -246,6 +247,14 @@ def Main() -> None:
                  'Force instance deletion when deletion protection is '
                  'activated.', False),
             ])
+  AddParser('aws', aws_subparsers, 'quarantinevm', 'Put a VM in '
+                                                   'network quarantine.',
+            args=[
+                ('instance_id', 'ID (i-xxxxxx) of the instance to quranitne.',
```
zkck, line 253:
```suggestion
                ('instance_id', 'ID (i-xxxxxx) of the instance to quarantine.',
```

zkck (commented):
LGTM, just the typo in cli.py

libcloudforensics/providers/aws/forensics.py
```diff
@@ -445,3 +446,36 @@ def CopyEBSSnapshotToS3(
   else:
     logger.info(
       'Image copy timeout. The process may be ongoing, or might have failed.')
+
+def InstanceNetworkQuarantine(zone: str,
```
giovannt0, line 450: New line before `zone`

libcloudforensics/providers/aws/forensics.py
```diff
@@ -445,3 +446,36 @@ def CopyEBSSnapshotToS3(
...
+
+  Args:
+    instance_id (str): : The id (i-xxxxxx) of the virtual machine.
+    exempted_src_subnets (List[str]): List of subnets that will be permitted
+  """
+  # Add /32 to any specified subnets that don't have a mask
+  # We're not checking the subnet is well formed, CreateIsolationSecurityGroup
+  # will take care of that
+  if exempted_src_subnets:
+    exempted_src_subnets[:] = [subnet if subnet.__contains__('/') else
```
giovannt0, line 467: Perhaps remove the call to `__contains__` and consider re-writing this as `[subnet if '/' in subnet else '{0:s}/32'.format(subnet) for subnet in exempted_src_subnets]`

libcloudforensics/providers/aws/forensics.py
```diff
@@ -445,3 +446,36 @@ def CopyEBSSnapshotToS3(
...
+      subnet + '/32' for subnet in exempted_src_subnets]
+
+  try:
+    aws_account = account.AWSAccount(zone)
```
giovannt0, line 471: This line can be moved out of the `try` block

libcloudforensics/providers/aws/forensics.py
```diff
@@ -445,3 +446,36 @@ def CopyEBSSnapshotToS3(
...
+    vpc = aws_account.ec2.GetInstanceById(instance_id).vpc
+    sg_id = \
+      aws_account.ec2.CreateIsolationSecurityGroup(vpc, exempted_src_subnets)
+    aws_account.ec2.SetInstanceSecurityGroup(instance_id, sg_id)
+  except errors.ResourceNotFoundError:
+    logger.error('Cannot qurantine non-existent instance')
```
giovannt0, line 477: For each of these errors, maybe we can log the stacktrace? Otherwise it's hard to understand what and why it went wrong

libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -566,3 +574,99 @@ def GenerateSSHKeyPair(self, vm_name: str) -> Tuple[str, str]:
...
+        Description=description,
+        GroupName=name,
+        VpcId=vpc)
+    except client.exceptions.ClientError as exception:
+      raise errors.ResourceCreationError(
+        'Could not create security group: {0!s}'.format(
+           exception), __name__) from exception
+
+    # SG's automatically include an everything outbound rule - remove it
+    sg_rules = client.describe_security_group_rules(
```
giovannt0, line 616: Do any of these calls (`clients.xxx`) throw an exception? If so, wrap them.

libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -566,3 +574,99 @@ def GenerateSSHKeyPair(self, vm_name: str) -> Tuple[str, str]:
               exception), __name__) from exception
     # If the call was successful, the response contains key information
     return key['KeyName'], key['KeyMaterial']
+
+  def CreateIsolationSecurityGroup(self,
```
giovannt0, line 578: Add a new line before `self`

libcloudforensics/providers/aws/internal/ec2.py
```diff
@@ -566,3 +574,99 @@ def GenerateSSHKeyPair(self, vm_name: str) -> Tuple[str, str]:
...
+          'IpProtocol': '-1',
+          'FromPort': 0,
+          'ToPort': 65535,
+          'IpRanges': ranges
+        }]
+      )
+
+    return str(sg_info['GroupId'])
+
+  def SetInstanceSecurityGroup(self,
```
giovannt0, line 653: Add a new line before `self`

giovannt0 (changes requested):
Few nits, thanks!

libcloudforensics/providers/aws/forensics.py
ramo-j: It can indeed, but I prefer the way it groups the code together from a style standpoint.

libcloudforensics/providers/aws/forensics.py
ramo-j: Looking at the other methods in `forensics.py` it looks to actually be standard to not catch exceptions at this level, and propagate them higher up the stack. So, adding a reraise for the instance not existing, and removing the other catches.

libcloudforensics/providers/aws/internal/ec2.py
ramo-j: My understanding is it will only throw a permission denied (which isn't considered anywhere else either) or a ResourceNotFound - but we just created the resource.

giovannt0: ACK, thanks for confirming

giovannt0 (approved):

PR #351, opened 2021-07-22
Remove external IPs when putting an instance into network quarantine (GCP)

### LINKED ISSUE #347: Remove external IP when in network quarantine
The current way we put GCE instances in network quarantine is via Firewall rules, this will prevent new connection but will not break on-going connections. In order to stop ongoing connections:
- we can either remove all external IPs after adding firewall rules then add them again
- remove all external IPs and add another jump instance to connect to the quarantined instance

zkck:
Closes #347 

The current way GCE instances are put into quarantine is via firewall rules. This prevents new connections, but does not break ongoing connections. Removing the external IP addresses will break these ongoing connections.

The additions in this PR are the following subsequent steps to quarantining the instance:
1. Removing all external IP addresses associated to an instance, breaking off all ongoing connections.
1. Restoring the external IPs. If the IP address was static, it will be re-applied. Otherwise if it was ephemeral, an ephemeral address will be re-applied.

### DIFF
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -440,14 +440,17 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
     region = '-'.join(zone_parts[:-1])
     # Request list of addresses
     addresses_client = self.GceApi().addresses()
-    params = dict(project=self.project_id, region=region)
+    params = {
+      'project': self.project_id, 
+      'region': region
+   }
     try:
       responses = common.ExecuteRequest(addresses_client, 'list', params)
     except HttpError as exception:
       message = 'Unable to list external IPs for {0:s}: {1:s}'.format(
         self.project_id,
         exception.error_details)
-      raise errors.ResourceNotFoundError(message, __name__)
+      raise errors.ResourceNotFoundError(message, __name__) from exception
     ip_addresses = []
     for response in responses:
       for address in response.get('items', []):
@@ -1026,13 +1029,13 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
     if ip_addr is not None:
       body['natIP'] = ip_addr
     instances_client = self.GceApi().instances()
-    params = dict(
-      project=self.project_id,
-      zone=self.zone,
-      instance=self.name,
-      networkInterface=net_if,
-      body=body
-    )
+    params = {
+      'project': self.project_id,
+      'zone': self.zone,
+      'instance': self.name,
+      'networkInterface': net_if,
+      'body': body
+    }
     try:
       # Safe to unpack, as the response is not paged
       response = common.ExecuteRequest(instances_client,
@@ -1042,7 +1045,7 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
       message = 'Unable to assign IP to {0:s}: {1:s}'.format(
         self.name,
         exception.error_details)
-      raise errors.ResourceCreationError(message, __name__)
+      raise errors.ResourceCreationError(message, __name__) from exception
     self.BlockOperation(response, self.zone)
 
   def RemoveExternalIps(self) -> Dict[str, str]:
@@ -1086,12 +1089,13 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
         ))
       # Execute the IP address removal by deleting access config
       gce_instance_client = self.GceApi().instances()
-      params = dict(
-        project=self.project_id,
-        zone=self.zone,
-        instance=self.name,
-        accessConfig=access_config_name,
-        networkInterface=network_interface_name)
+      params = {
+        'project': self.project_id,
+        'zone': self.zone,
+        'instance': self.name,
+        'accessConfig': access_config_name,
+        'networkInterface': network_interface_name
+      }
       try:
         # Safe to unpack since this response is not paged
         response = common.ExecuteRequest(gce_instance_client,
@@ -1101,7 +1105,7 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
         message = 'Unable to delete access config for {0:s}: {1:s}'.format(
           self.name,
           exception.error_details)
-        raise errors.ResourceDeletionError(message, __name__)
+        raise errors.ResourceDeletionError(message, __name__) from exception
       self.BlockOperation(response, zone=self.zone)
       # Save deleted external IP address
       external_ip_addresses[network_interface_name] = external_ip_address
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -1008,7 +1008,7 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
     Args:
       net_if (str): The instance's network interface to which the IP address
         must be assigned.
-      ip_addr (Optional[str]): The static IP address that exposes the network
+      ip_addr (str): Optional. The static IP address that exposes the network
         interface. If None, the assigned IP address will be ephemeral.
     """
     body = {}
@@ -1026,9 +1026,7 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
     self.BlockOperation(response, self.zone)
 
   def RemoveExternalIps(self) -> Dict[str, str]:
-    """
-    Removes any external IP of the instance, thus breaking
-    any ongoing connections.
+    """Removes any external IP of the instance, breaking ongoing connections.
 
     Note that if the instance's IP address was static, that
     the IP will still belong to the project.
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -963,6 +963,47 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
             self.FormatLogMessage(
                 'Could not find disk: {0:s}, skipping'.format(disk_name)))
 
+  def RemoveExternalIps(self) -> List[str]:
+    """
+    Removes any external IP of the instance, thus breaking
+    any ongoing connections
+    """
+    external_ip_addresses = []  # Result variable
+    # Iterate through instance's network interfaces, removing
+    # all access configurations (NAT)
+    instance_info = self.GetOperation()
+    for network_interface in instance_info['networkInterfaces']:
+      if 'accessConfigs' not in network_interface:
+        # No way to access this network interface externally,
+        # skip the removal
+        continue
+      # Extract relevant identifiers for deletion request
+      network_interface_name = network_interface['name']
+      access_config = network_interface['accessConfigs'][0]
+      access_config_name = access_config['name']
+      external_ip_address = access_config['natIP']
+      # Inform analyst of the deletion
+      logger.info(
+        "Deleting access config for {} (external IP: {})".format(
+          self.name,
+          external_ip_address,
+        ))
+      # Execute the IP address removal by deleting access
+      # config
+      gce_instance_client = self.GceApi().instances()
+      request = gce_instance_client.deleteAccessConfig(
+        project=self.project_id,
+        zone=self.zone,
+        instance=self.name,
+        accessConfig=access_config_name,
+        networkInterface=network_interface_name)
+      response = request.execute()
+      # Save deleted external IP address
+      self.BlockOperation(response, zone=self.zone)
+      external_ip_addresses.append(external_ip_address)
+    # Return the deleted external IP address for future use
+    return external_ip_addresses
+
   def SetTags(self, new_tags: List[str]) -> None:
     """Sets tags for the compute instance.
 
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -431,6 +431,7 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
 
     Raises:
       ValueError: If the zone is malformed.
+      errors.ResourceNotFoundError: If the request did not succeed.
     """
     # Convert zone to region
     zone_parts = zone.split('-')
@@ -440,7 +441,13 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
     # Request list of addresses
     addresses_client = self.GceApi().addresses()
     request = addresses_client.list(project=self.project_id, region=region)
-    response = request.execute()
+    try:
+      response = request.execute()
+    except HttpError as exception:
+      message = 'Unable to list external IPs for {0:s}: {1:s}'.format(
+        self.project_id,
+        exception.error_details)
+      raise errors.ResourceNotFoundError(message, __name__)
     ip_addresses = []
     for address in response.get('items', []):
       is_reserved = address['status'] == 'RESERVED'
@@ -1010,6 +1017,9 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
         must be assigned.
       ip_addr (str): Optional. The static IP address that exposes the network
         interface. If None, the assigned IP address will be ephemeral.
+
+    Raises:
+      errors.ResourceCreationError: If the assignment did not succeed.
     """
     body = {}
     if ip_addr is not None:
@@ -1022,7 +1032,13 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
       networkInterface=net_if,
       body=body
     )
-    response = request.execute()
+    try:
+      response = request.execute()
+    except HttpError as exception:
+      message = 'Unable to assign IP to {0:s}: {1:s}'.format(
+        self.name,
+        exception.error_details)
+      raise errors.ResourceCreationError(message, __name__)
     self.BlockOperation(response, self.zone)
 
   def RemoveExternalIps(self) -> Dict[str, str]:
@@ -1034,6 +1050,9 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
     Returns:
       Dict[str, str]: A mapping from an instance's network
         interfaces to the corresponding removed external IP.
+
+    Raises:
+      errors.ResourceDeletionError: If the removal did not succeed.
     """
     external_ip_addresses = {}
     # Iterate through instance's network interfaces, removing
@@ -1069,7 +1088,13 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
         instance=self.name,
         accessConfig=access_config_name,
         networkInterface=network_interface_name)
-      response = request.execute()
+      try:
+        response = request.execute()
+      except HttpError as exception:
+        message = 'Unable to delete access config for {0:s}: {1:s}'.format(
+          self.name,
+          exception.error_details)
+        raise errors.ResourceDeletionError(message, __name__)
       self.BlockOperation(response, zone=self.zone)
       # Save deleted external IP address
       external_ip_addresses[network_interface_name] = external_ip_address
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -417,9 +417,13 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
     return self._ListByLabel(
         labels_filter, disk_service_object, filter_union)
 
-  def ListReservedExternalIps(self, region: str) -> List[str]:
-    request = self.GceApi().addresses().list(project=self.project_id, region=region)
-    response = self.BlockOperation(request.execute())
+  def ListReservedExternalIps(self, zone: str) -> List[str]:
+    # Convert zone to region
+    region = '-'.join(zone.split('-')[:-1])
+    # Request list of addresses
+    addresses_client = self.GceApi().addresses()
+    request = addresses_client.list(project=self.project_id, region=region)
+    response = request.execute()
     addresses = []
     for address in response.get('items', []):
       is_reserved = address['status'] == 'RESERVED'
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -421,8 +421,8 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
     """Lists all static external IP addresses that are reserved (i.e. not in
     use) and can be used in a particular zone.
 
-    The method first converts the zone is converted to a region,
-    and then queries the GCE addresses.
+    The method first converts the zone to a region,
+    and then queries the GCE addresses resource.
 
     Args:
       zone (str): The zone in which the returned IPs would be available.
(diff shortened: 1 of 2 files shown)


libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -417,6 +417,37 @@ def ListDiskByLabels(self,
     return self._ListByLabel(
         labels_filter, disk_service_object, filter_union)
 
+  def ListReservedExternalIps(self, zone: str) -> List[str]:
+    """Lists all static external IP addresses that are reserved (i.e. not in
```
giovannt0, line 421: Mega-nit: I _think_ the first sentence of the docstring is supposed to hold on one line. To give additional details, do it below (in your `the method first converts .... ` block)

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+    """Assigns an external IP to an instance's network interface.
+
+    The instance must not have an IP assigned to the network interface when
+    calling this method. If the IP address is specified, it must be one that
+    is available to the project.
+
+    Args:
+      net_if (str): The instance's network interface to which the IP address
+        must be assigned.
+      ip_addr (Optional[str]): The static IP address that exposes the network
```
giovannt0, line 1007: In the docstring itself, you can write optional parameters as `param (str): Optional. [description]`

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+      instance=self.name,
+      networkInterface=net_if,
+      body=body
+    )
+    response = request.execute()
+    self.BlockOperation(response, self.zone)
+
+  def RemoveExternalIps(self) -> Dict[str, str]:
+    """
+    Removes any external IP of the instance, thus breaking
```
giovannt0, line 1026: `"""Remove....` on one line.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+    any ongoing connections.
+
+    Note that if the instance's IP address was static, that
+    the IP will still belong to the project.
+
+    Returns:
+      Dict[str, str]: A mapping from an instance's network
+        interfaces to the corresponding removed external IP.
+    """
+    external_ip_addresses = {}  # Result variable
```
giovannt0, line 1036: You can probably remove `# Result variable`, it's clear from context.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+    # Iterate through instance's network interfaces, removing
+    # all access configurations (NAT)
+    instance_info = self.GetOperation()
+    for network_interface in instance_info['networkInterfaces']:
```
giovannt0, line 1040: Perhaps change `instance_info['networkInterfaces']` to `instance_info.get('networkInterfaces', [])`, so we make sure not to have unexpected `KeyErrors` here

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+      if 'accessConfigs' not in network_interface:
+        # No way to access this network interface externally,
+        # skip the removal
+        continue
+      # Extract relevant identifiers for deletion request
+      network_interface_name = network_interface['name']
+      access_config = network_interface['accessConfigs'][0]
```
giovannt0, line 1047: I see you're accessing the first config in `network_interface['accessConfigs']`. Any scenario where there are more configs than just one? Adding a comment here to clarify why you expect the first element to be the one you need might be helpful.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+      access_config_name = access_config['name']
+      external_ip_address = access_config['natIP']
+      # Inform analyst of the deletion
```
giovannt0, line 1050: Remove this comment

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+      logger.info(
+        "Deleting access config for {} (external IP: {})".format(
```
giovannt0, line 1052: Add position and type to string format

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+      # Execute the IP address removal by deleting access
+      # config
+      gce_instance_client = self.GceApi().instances()
+      request = gce_instance_client.deleteAccessConfig(
+        project=self.project_id,
+        zone=self.zone,
+        instance=self.name,
+        accessConfig=access_config_name,
+        networkInterface=network_interface_name)
+      response = request.execute()
```
giovannt0, line 1065: Wrap this in a `try except` in case in fail; raise an error if it does and add that error to the docstring with an explanation of when it happens.

giovannt0 (changes requested):
Thanks Zak, great work! I was *extra* picky on purpose, please feel free to reach out if anything is unclear.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -417,6 +417,37 @@ def ListDiskByLabels(self,
...
+    """
+    # Convert zone to region
+    zone_parts = zone.split('-')
+    if len(zone_parts) != 3:
+      raise ValueError("Invalid zone: {}.")
+    region = '-'.join(zone_parts[:-1])
+    # Request list of addresses
+    addresses_client = self.GceApi().addresses()
+    request = addresses_client.list(project=self.project_id, region=region)
+    response = request.execute()
```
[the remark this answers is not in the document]
sa3eed3ed: You can use ExecuteRequest directly and it will handle retries and paging, it returns a list of responses (respnse pages) other than a single response, so you would also need to loop on it  https://github.com/google/cloud-forensics-utils/blob/a849b99e780726969fd3fecf8ccab1d4ddcc5ebd/libcloudforensics/providers/gcp/internal/common.py#L265

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
...
+      body['natIP'] = ip_addr
+    instances_client = self.GceApi().instances()
+    request = instances_client.addAccessConfig(
+      project=self.project_id,
+      zone=self.zone,
+      instance=self.name,
+      networkInterface=net_if,
+      body=body
+    )
+    response = request.execute()
```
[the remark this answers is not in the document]
sa3eed3ed: similarly also consider ExecuteRequest here to not add retries everywhere

libcloudforensics/providers/gcp/internal/compute.py
sa3eed3ed: similarly also consider ExecuteRequest here to not add retries everywhere, you will also need to loop on results and block on each response

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -417,6 +417,39 @@ def ListDiskByLabels(self,
...
+      raise ValueError('Invalid zone: {0:s}.'.format(zone))
+    region = '-'.join(zone_parts[:-1])
+    # Request list of addresses
+    addresses_client = self.GceApi().addresses()
+    request = addresses_client.list(project=self.project_id, region=region)
+    response = request.execute()
+    ip_addresses = []
+    for address in response.get('items', []):
+      is_reserved = address['status'] == 'RESERVED'
+      is_external = address['addressType'] == 'EXTERNAL'
```
sa3eed3ed, line 447: you return an empty list [] if items not exist, then you try to access dict fields in address['status'] and address['addressType'], in case address is [] this will fail.
```suggestion
    for address in response.get('items', {}):
      is_reserved = address.get('status') == 'RESERVED'
      is_external = address.get('addressType') == 'EXTERNAL'
```

giovannt0: I think here if `address` ends up being an empty list the `for` loop is skipped all-together

zkck: Yes, the body of the loop will not be executed in the case that `'items'` is not present.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -417,6 +417,37 @@ def ListDiskByLabels(self,
...
+    Args:
+      zone (str): The zone in which the returned IPs would be available.
+
+    Returns:
+      List[str]: The list of available IPs in the specified zone.
+    """
+    # Convert zone to region
+    zone_parts = zone.split('-')
+    if len(zone_parts) != 3:
+      raise ValueError("Invalid zone: {}.")
```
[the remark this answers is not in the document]
zkck: Thanks, I indeed missed the `.format()`.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
             self.FormatLogMessage(
                 'Could not find disk: {0:s}, skipping'.format(disk_name)))
 
+  def AssignExternalIp(self, net_if: str, ip_addr: Optional[str]) -> None:
```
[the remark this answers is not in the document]
zkck: Done, the intention was that the caller had to specify that he wanted an ephemeral IP by passing `None` as argument, but I understand how it's just extra code.

libcloudforensics/providers/gcp/internal/compute.py
zkck: Thanks for pointing this out, added clarification in the comments.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -417,6 +417,46 @@ def ListDiskByLabels(self,
...
+    # Request list of addresses
+    addresses_client = self.GceApi().addresses()
+    request = addresses_client.list(project=self.project_id, region=region)
+    try:
+      response = request.execute()
+    except HttpError as exception:
+      message = 'Unable to list external IPs for {0:s}: {1:s}'.format(
+        self.project_id,
+        exception.error_details)
+      raise errors.ResourceNotFoundError(message, __name__)
```
sa3eed3ed, line 450: better use ExecuteRequest that will handle errors and paging than adding try/except block for everu call, also paging isn't handled that way. Keep in mind that you will get a list of responses that you need to loop on https://github.com/google/cloud-forensics-utils/blob/a849b99e780726969fd3fecf8ccab1d4ddcc5ebd/libcloudforensics/providers/gcp/internal/common.py#L265

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1003,104 @@ def Delete(
...
+      body=body
+    )
+    try:
+      response = request.execute()
+    except HttpError as exception:
+      message = 'Unable to assign IP to {0:s}: {1:s}'.format(
+        self.name,
+        exception.error_details)
+      raise errors.ResourceCreationError(message, __name__)
+    self.BlockOperation(response, self.zone)
```
sa3eed3ed, line 1042: same as above use ExecuteRequest block on all returned responses

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1003,104 @@ def Delete(
...
+        accessConfig=access_config_name,
+        networkInterface=network_interface_name)
+      try:
+        response = request.execute()
+      except HttpError as exception:
+        message = 'Unable to delete access config for {0:s}: {1:s}'.format(
+          self.name,
+          exception.error_details)
+        raise errors.ResourceDeletionError(message, __name__)
+      self.BlockOperation(response, zone=self.zone)
```
sa3eed3ed, line 1098: same as above use ExecuteRequest block on all returned responses

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +994,81 @@ def Delete(
             self.FormatLogMessage(
                 'Could not find disk: {0:s}, skipping'.format(disk_name)))
 
+  def AssignExternalIp(self, net_if: str, ip_addr: Optional[str]) -> None:
+    """Assigns an external IP to an instance's network interface.
+
+    The instance must not have an IP assigned to the network interface when
```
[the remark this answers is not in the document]
sa3eed3ed: I think in the current scenario this wonn't happen since RemoveExternalIps removes all ips from all network interfaces and is just called before this in the call, but handling execute exceptions (maybe via ExecuteRequest) is for sure a good idea

sa3eed3ed (changes requested):
few small things, thank you for working on this

libcloudforensics/providers/gcp/internal/compute.py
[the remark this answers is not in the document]
zkck: Done, thanks for pointing it out. The API call is indeed a "paged" response, and ExecuteRequest handles that case 👍

libcloudforensics/providers/gcp/internal/compute.py
zkck: It seems that ExecuteRequest only handles errors with authentication; if there is any HttpError it will not be handled. It will instead be passed up to the caller so the try/catch block might still be necessary here.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+        project=self.project_id,
+        zone=self.zone,
+        instance=self.name,
+        accessConfig=access_config_name,
+        networkInterface=network_interface_name)
+      try:
+        # Safe to unpack since this response is not paged
+        response = common.ExecuteRequest(gce_instance_client,
+                                         'deleteAccessConfig',
+                                         params)[0]
```
zkck, line 1099: Concerning these sort of non-paged calls, I'd rather write it as 
```suggestion
        # Safe to unpack since this response is not paged
        response, = common.ExecuteRequest(gce_instance_client,
                                         'deleteAccessConfig',
                                         params)
```
since as per the API reference I **really** don't expect to see more than one response here, and if it is the case things should not just continue working as usual. If we write it as suggested then we get at least get an unpacking error when we get more than one response.

This suggestion also entails adding `unbalanced-tuple-unpacking` to the disabled list in pylintrc.

giovannt0: Imho there are a few problems with this approach: 
- It reduces readability (the `,` is not easy to spot, which makes the behaviour of this line hard to see). 
- Disabling `pylint` rules (especially at file level rather than line level) is generally speaking to be avoided, unless necessary. If you want to alert the user when the response contains more than one result, I'd suggest instead to log a warning:

```
responses = common.ExecuteRequest(gce_instance_client, 'deleteAccessConfig', params)
if len(responses) > 1:
  log.warning(....)
response = responses[0]  # Or access the member wherever used after in the code
```

API calls (their structures and responses) are bound to change over time. If a breaking change is introduced by the API provider, then our end-to-end tests are designed to catch them early.

libcloudforensics/providers/gcp/internal/compute.py
giovannt0: We might want to catch HttpError directly in `ExecuteRequest` in a separate block

zkck: Okay, so I think this calls for a couple enhancements to `ExecuteRequest`:

1. We catch the `HttpError` directly in the call, which would require an additional argument as to which exception to raise
2. We also need a way to do requests that only require a single response, and logs warnings in the case that there are more than one response

My suggestion would be to merge this PR, and then consider this refactoring a separate issue.

giovannt0: 1) Actually I _think_ we might have left out `HttpError` from `ExecuteRequest` voluntarily back then: catching the `HttpError` directly from the caller allows us to raise the proper error subsequently (e.g. here `ResourceNotFoundError`). So, scratch that; you can leave it as is, sorry for my earlier comment.

2) Imo it's okay to keep returning a list of responses, and handle the logic in the caller. `ExecuteRequest` is just a wrapper around executing the API call, and I don't think it should handle anything more than that (i.e. any logic on the responses themselves, including how many responses were returned or how they are formatted).

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -417,6 +417,47 @@ def ListDiskByLabels(self,
...
+      errors.ResourceNotFoundError: If the request did not succeed.
+    """
+    # Convert zone to region
+    zone_parts = zone.split('-')
+    if len(zone_parts) != 3:
+      raise ValueError('Invalid zone: {0:s}.'.format(zone))
+    region = '-'.join(zone_parts[:-1])
+    # Request list of addresses
+    addresses_client = self.GceApi().addresses()
+    params = dict(project=self.project_id, region=region)
```
giovannt0, line 443:
```suggestion
    params = {
      'project': self.project_id, 
      'region': region
   }
```
IIRC the python style guide discourages using `dict()` to construct dictionaries.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -417,6 +417,47 @@ def ListDiskByLabels(self,
...
+    try:
+      responses = common.ExecuteRequest(addresses_client, 'list', params)
+    except HttpError as exception:
+      message = 'Unable to list external IPs for {0:s}: {1:s}'.format(
+        self.project_id,
+        exception.error_details)
+      raise errors.ResourceNotFoundError(message, __name__)
```
giovannt0, line 450:
```suggestion
      raise errors.ResourceNotFoundError(message, __name__) from exception
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+    try:
+      # Safe to unpack, as the response is not paged
+      response = common.ExecuteRequest(instances_client,
+                                       'addAccessConfig',
+                                       params)[0]
+    except HttpError as exception:
+      message = 'Unable to assign IP to {0:s}: {1:s}'.format(
+        self.name,
+        exception.error_details)
+      raise errors.ResourceCreationError(message, __name__)
```
giovannt0, line 1045:
```suggestion
      raise errors.ResourceCreationError(message, __name__) from exception
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+        interface. If None, the assigned IP address will be ephemeral.
+
+    Raises:
+      errors.ResourceCreationError: If the assignment did not succeed.
+    """
+    body = {}
+    if ip_addr is not None:
+      body['natIP'] = ip_addr
+    instances_client = self.GceApi().instances()
+    params = dict(
```
giovannt0, line 1029:
```suggestion
    params = {
      'project': self.project_id,
      'zone': self.zone,
      'instance': self.name,
      'networkInterface': net_if,
      'body': body
    }
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+      project=self.project_id,
```
giovannt0, line 1030:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+      zone=self.zone,
```
giovannt0, line 1031:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+      instance=self.name,
```
giovannt0, line 1032:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+      networkInterface=net_if,
```
giovannt0, line 1033:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+      body=body
```
giovannt0, line 1034:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+    )
```
giovannt0, line 1035:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+      access_config_name = access_config['name']
+      external_ip_address = access_config['natIP']
+      logger.info(
+        'Deleting access config for {0:s} (external IP: {1:s})'.format(
+          self.name,
+          external_ip_address,
+        ))
+      # Execute the IP address removal by deleting access config
+      gce_instance_client = self.GceApi().instances()
+      params = dict(
```
giovannt0, line 1089:
```suggestion
      params = {
        'project': self.project_id,
        'zone': self.zone,
        'instance': self.name,
        'accessConfig': access_config_name,
        'networkInterface': network_interface_name
      }
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+        project=self.project_id,
```
giovannt0, line 1090:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+        zone=self.zone,
```
giovannt0, line 1091:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+        instance=self.name,
```
giovannt0, line 1092:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+        accessConfig=access_config_name,
```
giovannt0, line 1093:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+        networkInterface=network_interface_name)
```
giovannt0, line 1094:
```suggestion
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -963,6 +1004,110 @@ def Delete(
...
+      try:
+        # Safe to unpack since this response is not paged
+        response = common.ExecuteRequest(gce_instance_client,
+                                         'deleteAccessConfig',
+                                         params)[0]
+      except HttpError as exception:
+        message = 'Unable to delete access config for {0:s}: {1:s}'.format(
+          self.name,
+          exception.error_details)
+        raise errors.ResourceDeletionError(message, __name__)
```
giovannt0, line 1104:
```suggestion
        raise errors.ResourceDeletionError(message, __name__) from exception
```

giovannt0 (approved):
Left a few suggestions; LGTM. Thanks!

sa3eed3ed (approved):
LGTM

PR #358, opened 2021-07-28
Adjusted ebs snapshot s3 copy to wait for the instance health check

ramo-j:
This allows more wait time for the instance to start before checking for success starts.


GreenStorm-Code (approved):
Looks good to me and premise makes sense

Issue #339, opened 2021-07-12
Add functionality to list firewall rules attached to a compute instance


GreenStorm-Code:
Showing the firewall rules can easily give an assessment of why an instance could have been popped

PR #361, opened 2021-08-02
GCP effective firewall listing

jonathan-greig:
Relates to #339

### DIFF
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -1075,6 +1075,80 @@ class GoogleComputeInstance(compute_base_resource.GoogleComputeBaseResource):
       raise errors.ServiceAccountRemovalError('Service account detatchment '
           'failure: {0:s}'.format(str(exception)), __name__)
 
+  def GetNetworkInterfaces(self) -> List[Any]:
+    """Get the network interfaces for an instance.
+
+    Returns:
+      List[Any]: A list of network interfaces in the format documented at
+        https://cloud.google.com/compute/docs/reference/rest/v1/instances/get 
+    """
+
+    get_operation = self.GetOperation()
+    return get_operation.get('networkInterfaces')
+
+  def _NormaliseFirewallRules(self, nic_rules: Dict[str, Any]) -> List[Any]:
+    """Normalise firewall policies and firewall rules into a common format.
+
+    Args:
+      nic_rules: the effective firewall rules for an individual NIC.
+
+    Returns:
+      The normalised firewall rules.
+    """
+    normalised_rules = []
+    firewall_policies = nic_rules['firewallPolicys']
+    firewalls = nic_rules['firewalls']
+
+    for policy_level, policy in enumerate(firewall_policies):
+      for rule in policy['rules']:
+        match = rule['match']
+        ingress = rule['direction'] == 'INGRESS'
+        normalised_rule = {
+            'type': 'policy',
+            'policy_level': policy_level,
+            'priority': rule['priority'],
+            'direction': rule['direction'],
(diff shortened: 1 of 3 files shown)


tests/providers/gcp/gcp_mocks.py
```diff
@@ -556,3 +556,97 @@
...
+            {
+                'type': 'ONE_TO_ONE_NAT',
+                'name': 'External NAT',
+                'natIP': '0.0.0.0',
+                'networkTier': 'PREMIUM',
+                'kind': 'compute#accessConfig'
+                }
+        ],
+        'fingerprint': 'bm9mcGZwZnA=',
+        'kind': 'compute#networkInterface'}
```
giovannt0, line 577: Nit: put the bracket on the next line so we clearly see where the dict finishes

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -1224,6 +1224,85 @@ def DetachServiceAccount(self) -> None:
...
+    gce_instance_client = self.GceApi().instances()
+    instance_info = self.GetOperation()
+    fw_rules = {}
+
+    for nic in instance_info.get('networkInterfaces', []):
+      nic_name = nic['name']
+      request = gce_instance_client.getEffectiveFirewalls(
+          project=self.project_id, instance=self.name, zone=self.zone,
+          networkInterface=nic_name)
+      response = request.execute()
```
giovannt0, line 1300: You may wanna wrap this using `common.ExecuteRequest`, so that errors are caught.

giovannt0 (changes requested):

jonathan-greig:
Addressed code review comments and also did some further normalisation of the fw rules as there were inconsistencies in FW vs FW policy field names `IPProtocol` vs `ipProtocol`

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -1224,6 +1226,112 @@ def DetachServiceAccount(self) -> None:
...
+    gce_instance_client = self.GceApi().instances()
+    instance_info = self.GetOperation()
+    fw_rules = {}
+
+    for nic in instance_info.get('networkInterfaces', []):
+      nic_name = nic['name']
+      nic_fw_rules = []
+      responses = common.ExecuteRequest(
+          gce_instance_client, 'getEffectiveFirewalls',
+          {'project': self.project_id, 'instance': self.name,
```
giovannt0, line 1326: Can you make this a separate variable? `request = {...}`

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -1224,6 +1226,112 @@ def DetachServiceAccount(self) -> None:
...
+
+    Returns:
+      List[Any]: the normalised l4config"""
+    normalised_l4config = []
+    for config in l4config:
+      normalised = {}
+      if 'ipProtocol' in config:
+        normalised['ip_protocol'] = config['ipProtocol']
+      else:
+        normalised['ip_protocol'] = config['IPProtocol']
```
giovannt0, line 1244: If `ipProtocol` is not in `config`, is it always guaranteed that `IPProtocol` is?

giovannt0 (changes requested):

libcloudforensics/providers/gcp/internal/compute.py
jonathan-greig: Yeah according to the documentation each "L4config" will have either an `ipProtocol` if it is a firewall policy or `IPProtocol` if it is a standard firewall rule. Not sure why they are named differently :smiley: . `elif` is probably more suitable here though will change.

giovannt0 (approved):

PR #368, opened 2021-08-04
Method to abandon instance from managed instance group

zkck:
When quarantining a GKE workload, we first want to abandon the affected nodes from their managed instance group so that they don't get marked as unhealthy and auto-repaired.


libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -1112,6 +1112,48 @@ def RemoveExternalIps(self) -> Dict[str, str]:
...
+    """
+
+    def RaiseException(exception: Exception, details: str) -> None:
+      msg = ('Unable to abandon {0:s} from managed '
+             'instance group {1:s}: {2:s}').format(
+        self.name,
+        instance_group,
+        details,
+      )
+      raise errors.ResourceDeletionError(msg, __name__) from exception
```
sa3eed3ed, line 1133: why did you decide to wrap the error in ResourceDeletionError? My thoughts are if the error is due to some lacking permission or resource not found (wrong name) this will make the error message can be confusing

zkck: I agree, I'll change it to `ResourceNotFoundError`.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -1112,6 +1112,48 @@ def RemoveExternalIps(self) -> Dict[str, str]:
...
+    """
+
+    def RaiseException(exception: Exception, details: str) -> None:
+      msg = ('Unable to abandon {0:s} from managed '
+             'instance group {1:s}: {2:s}').format(
+        self.name,
+        instance_group,
+        details,
+      )
+      raise errors.ResourceNotFoundError(msg, __name__) from exception
```
sa3eed3ed, line 1133: maybe you can just log and leave the original exception here or add a new generic exception type "ApiOperationFailed" and wrap any exception in it.
Because if you make it a specific exception type, you should have conditional clauses on the http and runtime errors you catch below to check the error code and then decide to raise not found or resource deleted or permission denied.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -1112,6 +1112,48 @@ def RemoveExternalIps(self) -> Dict[str, str]:
...
 
+  def AbandonFromMIG(self, instance_group: str) -> None:
+    """Abandons the instance from the managed instance group.
+
+    Args:
+      instance_group (str): The instance group that this instance should
+          be abondoned from.
+
+    Raises:
+      errors.ResourceDeletionError: If the request did not succeed.
```
sa3eed3ed, line 1123: Also don't forget to change this

sa3eed3ed (changes requested):
few nits

libcloudforensics/providers/gcp/internal/compute.py
zkck: Thanks, I opted for a more generic exception type.

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -1112,6 +1112,48 @@ def RemoveExternalIps(self) -> Dict[str, str]:
...
+
+    Args:
+      instance_group (str): The instance group that this instance should
+          be abondoned from.
+
+    Raises:
+      errors.OperationFailedError: If the request did not succeed.
+    """
+
+    def RaiseException(exception: Exception, details: str) -> None:
```
giovannt0, line 1126: Suggestion:

I see you're calling this method in two ways:
- `RaiseException(exception, exception.error_details)`
- `RaiseException(exception, str(exception))`

It looks like it would be enough to have a single argument in `RaiseException`: `RaiseException(exception: Exception)`

You can then pass the exception to the formatted string (you can use `{2!s}` to cast `exception` to a string, or pass `exception.error_details` if that's already a string).

giovannt0, line 1126: Suggestion:

I see you're calling this method in two ways:
- `RaiseException(exception, exception.error_details)`
- `RaiseException(exception, str(exception))`

It looks like it would be enough to have a single argument in `RaiseException`: `RaiseException(exception: Exception)`

You can then pass the exception to the formatted string (you can use `{2!s}` to cast `exception` to a string, or pass `exception.error_details` if that's already a string).

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -1112,6 +1112,48 @@ def RemoveExternalIps(self) -> Dict[str, str]:
...
+      raise errors.OperationFailedError(msg, __name__) from exception
+
+    mig_client = self.GceApi().instanceGroupManagers()
+    params = {
+      'project': self.project_id,
+      'zone': self.zone,
+      'instanceGroupManager': instance_group,
+      'body': {
+        'instances': [
+          f'zones/{self.zone}/instances/{self.name}'
```
giovannt0, line 1142: `'....'.format()` for consistency

libcloudforensics/providers/gcp/internal/compute.py
zkck: So I'd check if `error_details` is an attribute of the object, if so, pass it to the format string, else pass the cast of the exception?

giovannt0: I'd just directly pass `exception` to the formatted string: 
```python
msg = ('Unable to abandon {0:s} from managed instance group {1:s}: {2!s}').format(
  self.name,
  instance_group,
  exception
)
```

giovannt0: Actually, since you `raise from` it's ok to not format the exception in the string (so only keep `self.name` and `instance_group`)

zkck: Okay sounds good

giovannt0:
Thanks Zack, I'll let @sa3eed3ed give the final approval in case there's more stuff on his side.

sa3eed3ed (approved):
LGTM

PR #371, opened 2021-08-09
Break up CopyEBSSnapshotToS3

### LINKED ISSUE #370: Breakup ebs snapshot copy into multiple methods
Currently there is a single method `aws.forensics.CopyEBSSnapshotToS3` which covers a lot of actions. In addition to general readability, this should be broken up into a set up, run and tear down for parallelisation. 

Consider the scenario where a implementation wants to copy multiple snapshots in parallel. Multiple invocations will each try to create the IAM components if they do not exist already. If a pause is added between invocations to account for this, the first invocation will create the IAm components, and latter ones will see the IAM roles exist - However, the pause needs to be long enough for EC2 to pick up that the role is available as an instance role.

ramo-j:
Refactor CopyEBSSnapshotToS3 into 3 methods (SetUp, Process and TearDown) to better allow for parallel processing of snapshots. Also:

* Moved the relevant E2E test to an independent test class. 
* Rectify an small oversight from earlier PR - Add gcp storagetransfer module into gcp project. 

Closes #370.

### DIFF
--- a/libcloudforensics/providers/aws/forensics.py
+++ b/libcloudforensics/providers/aws/forensics.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Forensics on AWS."""
-from typing import TYPE_CHECKING, Tuple, List, Optional, Dict
+from typing import TYPE_CHECKING, Tuple, List, Optional, Dict, Any
 
 import random
 from time import sleep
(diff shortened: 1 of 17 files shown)


libcloudforensics/providers/aws/forensics.py
```diff
@@ -308,15 +308,63 @@ def StartAnalysisVm(
   return analysis_vm, created
 # pylint: enable=too-many-arguments
 
-def CopyEBSSnapshotToS3(
+def CopyEBSSnapshotToS3SetUp(
+    aws_account: account.AWSAccount,
+    instance_profile_name: str) -> Dict[str, Dict[str, Any]]:
+  """Set up for CopyEBSSnapshotToS3. Creates the IAM components required, or
+  returns the existing if they exist already.
```
Fryyyyy, line 315:
```suggestion
  returns the existing ones if they exist already.
```

Fryyyyy (approved):
Tiny comment nit otherwise LGTM

PR #374, opened 2021-08-13
Add cloud resource manager functionality

### LINKED ISSUE #373: Add functionality to list project hierarchy

jonathan-greig:
Closes #373


libcloudforensics/providers/gcp/internal/cloudresourcemanager.py
```diff
@@ -0,0 +1,87 @@
...
+      List[Any]: the list of ancestor resources.
+    """
+
+    ancestry = []
+    current_resource = self.GetResource('projects/' + project_number)
+    ancestry.append(current_resource)
+
+    while 'parent' in current_resource:
+      parent_name = current_resource['parent']
+      current_resource = self.GetResource(parent_name)
```
ramo-j, line 57: You have the method `GetResource` annotated with `Raises: TypeError` - Should we wrap calls to it in a try, or annotate this method with the same `Raises`?

ramo-j (commented):
Don't forget to add new internal components to the project class in `libcloudforensics/providers/gcp/internal/project.py`.

Otherwise, LGTM

ramo-j (approved):

libcloudforensics/providers/gcp/internal/cloudresourcemanager.py
jonathan-greig: Was wondering about that but the way `GetResource` is called here ensures that `projects/` is already prepended (line 52 above) and so should never end up raising that exception. I included the exception more for any other code that ends up calling this function and calls it with improperly formatted resource names.

PR #398, opened 2021-09-23
GKE integrate NetworkPolicy check to quarantining process

zkck:
Primary reason for this PR is integrating a NetworkPolicy check to the quarantining process, i.e. disabling the `Isolate pods?` option appropriately.

Additionally, this PR adds a few small changes:
- To disable a `PromptOption`, a reason for disabling is now required
- Fix a bug in which a workload's nodes may appear twice, by introducing a `K8sWorkload.GetCoveredNodes` method
- Add workaround to pod isolation when deleting workload
- Add appropriate defaults to the yes/no prompt options
- Fix strikethrough function
- Fix wrong string argument

### DIFF
--- a/libcloudforensics/prompts.py
+++ b/libcloudforensics/prompts.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """Classes for displaying prompts to a user and running their choices."""
 import abc
-from typing import List, Optional, Callable
+from typing import List, Optional, Callable, Tuple
 
 from libcloudforensics import logging_utils
 
@@ -31,7 +31,7 @@ def _Strikethrough(text: str) -> str:
   Returns:
     str: The given text with strikethrough codes.
   """
-  return ''.join('{0:s}\u0336'.format(char) for char in text)
+  return ''.join('\u0336{0:s}'.format(char) for char in text)
 
 
 class PromptOption:
@@ -45,7 +45,7 @@ class PromptOption:
       self,
       text: str,
       *functions: Callable[[], None],
-      disable_options: Optional[List['PromptOption']] = None) -> None:
+      disable_options: Optional[List[Tuple['PromptOption', str]]] = None) -> None:
     """Builds a PromptOption.
 
     Args:
@@ -57,7 +57,7 @@ class PromptOption:
     """
     self._text = text
     self._functions = functions
-    self._disabled = False
+    self._disabled_reason = None
     self._selected = False
     self._disable_options = disable_options or []
 
@@ -66,9 +66,9 @@ class PromptOption:
     """The text description of this PromptOption."""
     return self._text
 
-  def Disable(self) -> None:
+  def Disable(self, reason: str) -> None:
     """Disables this prompt."""
-    self._disabled = True
+    self._disabled_reason = reason
 
   def IsDisabled(self) -> bool:
     """Returns True if this prompt is disabled, false otherwise.
@@ -76,12 +76,12 @@ class PromptOption:
     Returns:
       bool: True if this prompt is disabled, false otherwise.
     """
-    return self._disabled
+    return self._disabled_reason is not None
 
   def Select(self) -> None:
     """Selects this prompt, disabling dependent prompts."""
-    for option in self._disable_options:
-      option.Disable()
+    for option, reason in self._disable_options:
+      option.Disable(reason)
     self._selected = True
 
   def IsSelected(self) -> bool:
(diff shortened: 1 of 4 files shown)


libcloudforensics/prompts.py
```diff
@@ -66,22 +69,26 @@ def text(self) -> str:
...
       bool: True if this prompt is disabled, false otherwise.
     """
-    return self._disabled
+    return self._disabled_reason is not None
 
   def Select(self) -> None:
     """Selects this prompt, disabling dependent prompts."""
-    for option in self._disable_options:
-      option.Disable()
+    for option, reason in self._disable_options:
```
giovannt0, line 90: `self._disable_options` is a list of Tuple, so I expect you're going to have a `TypeError` here.
You want to first iterate on that list, and then unpack the tuple into `option` and `reason`
```python
for disable in self._disable_options:
  for option, reason in disable:
    option.Disable(reason)
```

libcloudforensics/providers/kubernetes/mitigation.py
```diff
@@ -30,40 +32,49 @@ def DrainWorkloadNodesFromOtherPods(
...
 
-def CreateDenyAllNetworkPolicyForWorkload(
+def IsolatePodsWithNetworkPolicy(
     cluster: k8s.K8sCluster,
-    workload: workloads.K8sWorkload) -> netpol.K8sDenyAllNetworkPolicy:
-  """Isolates a workload's pods via a deny all network policy.
+    pods: List[base.K8sPod]) -> Optional[netpol.K8sDenyAllNetworkPolicy]:
+  """Isolates pods via a deny-all NetworkPolicy.
+
+  **Warning:** It is the caller's responsibility to make sure that Kubernetes
```
giovannt0, line 48: This should probably be logged as a warning in the method, because I don't expect people to read this (esp. those that are using the lib through `pip install`

libcloudforensics/providers/kubernetes/mitigation.py
```diff
@@ -30,40 +32,49 @@ def DrainWorkloadNodesFromOtherPods(
...
-    netpol.K8sDenyAllNetworkPolicy: The deny all network policy that was
-        created to isolate the workload's pods.
+    netpol.K8sDenyAllNetworkPolicy: Optional. The deny-all network policy that
+        was created to isolate the pods. If no pods were supplied, None is
+        returned.
+
+  Raises:
+    ValueError: If the pods are not in the same namespace.
   """
-  # TODO: Check that network policies are enabled
```
giovannt0, line 57: Is there a reason why we don't check and error if it's not, instead of putting this responsibility on the user?

giovannt0 (changes requested):

libcloudforensics/prompts.py
zkck: Values of a list can be unpacked directly with the variables given in the for loop: 

```
>>> l = [(0, 'foo'), (1, 'bar')]
>>> for i, val in l:
...     print(i, val)
... 
0 foo
1 bar
>>> for i, val in enumerate(['a', 'b', 'c']):
...     print(i, val)
... 
0 a
1 b
2 c

```

If I try out the code snippet:

```
>>> for tup in l:
...     for i, val in tup:
...         print(i, val)
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
TypeError: cannot unpack non-iterable int object
```

giovannt0: Not sure what I was thinking when I wrote this.... :) sorry about that, ignore me ;) Thanks for flagging

libcloudforensics/providers/kubernetes/mitigation.py
zkck: Yes, the intention was to make this `kubernetes` package standalone, and not rely on any other APIs than the Kubernetes API. If we want checks to be integrated, then we can move these functions to the GCP package. The problem I'd see with this is that it may lead to code duplication when we support AWS/Azure clusters.

zkck: Okay so a solution that may work is making the `K8sCluster` abstract, and have an abstract method that checks if NetworkPolicy is enabled, and have the cloud provider packages subclass it. We can then check if it is enabled in this function.

There is already a class restructuring branch that I have, so I'll add this change there, in a separate PR. I've made an issue for this #400

giovannt0 (approved):

PR #402, opened 2021-10-06
Kubernetes class restructure

zkck:
- Move the `K8sWorkload` abstract class higher up the hierarchy so that `K8sPod` can inherit from it, old `K8sWorkload` becomes `K8sControlledWorkload`
- Make `K8sCluster` an abstract class to be subclassed in the cloud provider packages (making `GkeCluster.GetK8sCluster` obsolete)
- Integrate NetworkPolicy check directly in isolation function
- Remove now unused method `AddTemplateLabels`
- Add a couple extra methods to `K8sCluster`
- Make `K8sDeployment.OrphanPods` more stable by saving the replica set before deletion

### DIFF
--- a/libcloudforensics/providers/gcp/forensics.py
+++ b/libcloudforensics/providers/gcp/forensics.py
@@ -501,24 +501,22 @@ def QuarantineGKEWorkload(project_id: str,
     namespace (str): The Kubernetes namespace of the workload (e.g. 'default').
     workload_id (str): The name of the workload.
   """
-  gke_cluster = gke.GkeCluster(project_id, zone, cluster_id)
-  k8s_cluster = gke_cluster.GetK8sCluster()
-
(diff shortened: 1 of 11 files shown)


libcloudforensics/providers/gcp/forensics.py
```diff
@@ -541,18 +539,18 @@ def IsolatePods() -> None:
     logger.info(
         'Creating deny-all NetworkPolicy for {0:s} '
         'workload...'.format(workload_id))
-    mitigation.IsolatePodsWithNetworkPolicy(k8s_cluster, workload_pods)
+    mitigation.IsolatePodsWithNetworkPolicy(cluster, workload_pods)
```
sa3eed3ed, line 542: if network policy is not enabled this will fail (raise error), do you check if network policy is enabled before calling IsolatePods, or maybe disable the option of IsolatingPod if network policy isn't enabled, to avoid failing

PR #404, opened 2021-10-08
FindWorkload method for K8sCluster

zkck:
This PR adds a method to find a workload in a Kubernetes cluster by name and namespace, together with its supporting methods.

##### `K8sControlledWorkload.OrphanPods`

This method needed to be moved higher up the hierarchy, to `K8sWorkload`, to be compatible with the returned type of `FindWorkload`. This meant having to implement it in `K8sPod`.

I chose that `OrphanPods` have no effect in `K8sPod` as opposed to deleting the pod. The reasoning for this decision was that the analyst may choose the `Preserve evidence and delete workload` prompt option with a pod workload specified, and if `OrphanPods` deletes the pod, then it will be deleting evidence, contradicting the prompt option. I added a warning to notify the user about this behavior.

### DIFF
--- a/tests/providers/azure/internal/test_compute.py
+++ b/tests/providers/azure/internal/test_compute.py
@@ -115,7 +115,7 @@ class AZComputeTest(unittest.TestCase):
     self.assertEqual('fake-region', disk.region)
     self.assertEqual(['fake-zone'], disk.zones)
 
-  @mock.patch('azure.mgmt.compute.v2020_12_01.operations._disks_operations.DisksOperations.begin_create_or_update')
+  @mock.patch('azure.mgmt.compute.v2021_04_01.operations._disks_operations.DisksOperations.begin_create_or_update')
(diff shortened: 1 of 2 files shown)


libcloudforensics/providers/kubernetes/cluster.py
```diff
@@ -128,6 +171,43 @@ def _AuthorizationCheck(self) -> None:
           'This object\'s client is not authorized to perform all operations'
           'on the Kubernetes cluster. API calls may fail.')
 
+  def AllWorkloads(self,
+                   namespace: Optional[str]) -> Iterable[base.K8sWorkload]:
```
sa3eed3ed, line 175:
```suggestion
                   namespace: Optional[str]) -> Iterable[List[base.K8sWorkload]]:
```

sa3eed3ed (changes requested):
few things

PR #419, opened 2021-11-04
Add owner label to GCP vms

ramo-j:
What is says on the tin, really. Owner is sourced as the current logged in user. Preferable would be the user authenticated to GCP but I have been unable to identify a method for discovering that.

But also, a small change for linter appeasement.

### DIFF
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -673,9 +673,13 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
 
     # TODO(ramo-j) Better than OS login would be the user authenticated to gcp,
     # but I cannt find a way to determine that.
-    if owner_label and os.getlogin():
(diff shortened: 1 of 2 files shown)


libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -667,6 +670,16 @@ def CreateInstanceFromArguments(  #pylint: disable=too-many-arguments
         }]
     }
     request_body['networkInterfaces'] = [network_interface]
+
+    # TODO(ramo-j) Better than OS login would be the user authenticated to gcp,
+    # but I cannt find a way to determine that.
+    try:
+      if owner_label and os.getlogin():
+        owner = re.sub('[^-_a-z0-9]', '_', os.getlogin())
+        request_body['labels'] = [{'owner':owner}]
```
tomchop, line 679:
```suggestion
        request_body['labels'] = [{'owner': owner}]
```

tomchop (approved):
One minor nit, otherwise LGTM!

PR #422, opened 2021-11-27
[azure] add image_reference option, fix multi-subscriptions bug, docs

juju4:
* adding image_reference option: this allows to use a different sku than default one (bionic) and also to use a shared gallery image with any custom image
* fix multi-subscription bug: previous state was checking existing of resource_group_name in src_profile which works fine if src and dst subscriptions are the same but not if different
* docs: more details especially on app token creation, azure permissions, and image_reference

### DIFF
--- a/docs/usermanual/index.md
+++ b/docs/usermanual/index.md
@@ -318,6 +318,7 @@ of in a different account (Account B).
 #### Using the library and the CLI
 
 To make the copy, import the `forensics` package and use the `CreateDiskCopy` method.
+`resource_group_name` is the destination resource group where disk copy and eventually analysis VM will be created.
 
 ```python
 from libcloudforensics.providers.azure import forensics
@@ -404,6 +405,12 @@ cloudforensics az 'resource_group_name' copydisk --disk_name='disk1' --src_profi
 # dst_profile, and attach a volume to it.
 # A SSH key pair will be automatically generated and associated to the instance.
 cloudforensics az 'resource_group_name' startvm 'vm-forensics' --disk_size=50 --cpu_cores=4 --attach_disks='disk1-copy' --dst_profile='dst_profile'
+
+# Start an analysis VM from a different publisher/sku
+cloudforensics az 'resource_group_name' startvm 'vm-forensics' --cpu_cores=4 --attach_disks='disk1-copy' --dst_profile='dst_profile' --image_reference='{ "publisher": "Canonical","offer": "0001-com-ubuntu-server-focal","sku": "20_04-lts","version": "latest"}'
+
+# Start an analysis VM from a shared gallery image, like pre-built forensics image
+cloudforensics az 'resource_group_name' startvm 'vm-forensics' --cpu_cores=4 --attach_disks='disk1-copy' --dst_profile='dst_profile' --image_reference='{"id": "/subscriptions/<GUID>/resourceGroups/Testing/providers/Microsoft.Compute/galleries/<mygallery>/images/<myimagename>/versions/latest"}'
 ```
 
 You're now ready to go! Log in your Azure account, find the instance's IP address and SSH to it.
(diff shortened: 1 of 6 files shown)


giovannt0:
Thank you for your contribution! Please have a look at https://cla.developers.google.com/

docs/usermanual/index.md
```diff
@@ -404,11 +405,69 @@ cloudforensics az 'resource_group_name' copydisk --disk_name='disk1' --src_profi
...
 You're now ready to go! Log in your Azure account, find the instance's IP address and SSH to it.
 
 ``` important:: Pro tip: you can export an environment variable 'STARTUP_SCRIPT' that points to a custom bash script. 
 This script will be shipped to the instance being created and executed during the first boot. You can do any kind of 
 pre-processing you want in this script.
+```
+
+### Azure App token creation
+
+You can do either through portal, either through following command lines (with azure cloudshell for example):
```
giovannt0, line 425:
```suggestion
You can do this either through Azure portal, or through the following command lines (with azure cloudshell for example):
```

libcloudforensics/providers/azure/forensics.py
```diff
@@ -164,6 +170,9 @@ def StartAnalysisVm(
     region (str): Optional. The region in which to create the VM.
         Default is eastus.
     attach_disks (List[str]): Optional. List of disk names to attach to the VM.
+    image_reference (Dict[str, str]): Optional. A dictionary of Azure image
+        to bootstrap. It can be usual sku/publisher/version/offer or shared
```
giovannt0, line 174:
```suggestion
        to bootstrap. It can be a sku/publisher/version/offer or a shared
```

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -66,8 +66,11 @@ def __init__(self,
     self._network = None  # type: Optional[network_module.AZNetwork]
     self._resource = None  # type: Optional[resource_module.AZResource]
     self._storage = None  # type: Optional[storage_module.AZStorage]
-    self.default_resource_group_name = self.resource.GetOrCreateResourceGroup(
-        default_resource_group_name)
+    if default_resource_group_name:
+      self.default_resource_group_name = self.resource.GetOrCreateResourceGroup(
+          default_resource_group_name)
+    else:
+      self.default_resource_group_name = ''
```
giovannt0, line 73: I think here what we could do is make `GetOrCreateResourceGroup` return `''` if the given resource name is `''`. This way, you can rewrite `init()` as:

```python
  def __init__(self,
               default_resource_group_name: str = '',
               default_region: str = 'eastus',
               profile_name: Optional[str] = None) -> None:
```

And assign the default resource name as:

```python
self.default_resource_group_name = self.resource.GetOrCreateResourceGroup(default_resource_group_name)
```

Without the `if/else` block. You just need to add the following block to `GetOrCreateResourceGroup` at its beginning:

```python
if not resource_group_name:
  return ''
```

libcloudforensics/providers/azure/internal/compute.py
```diff
@@ -393,6 +394,9 @@ def GetOrCreateAnalysisVm(
           provided, the vm will be created in the default_region
           associated to the AZAccount object.
       packages (List[str]): Optional. List of packages to install in the VM.
+      image_reference (Dict[str, str]): Optional. A dictionary of Azure image
+          to bootstrap. It can be usual sku/publisher/version/offer or shared
```
giovannt0, line 398:
```suggestion
          to bootstrap. It can be a sku/publisher/version/offer or a shared
```

libcloudforensics/providers/azure/internal/compute.py
```diff
@@ -432,16 +436,20 @@ def GetOrCreateAnalysisVm(
     if not region:
       region = self.az_account.default_region
 
+    if not image_reference:
+      image_reference = {
+                    'sku': common.UBUNTU_1804_SKU,
+                    'publisher': 'Canonical',
+                    'version': 'latest',
+                    'offer': 'UbuntuServer'}
```
giovannt0, line 444:
```suggestion
      image_reference = {
        'sku': common.UBUNTU_1804_SKU,
        'publisher': 'Canonical',
        'version': 'latest',
        'offer': 'UbuntuServer'
      }
```

libcloudforensics/providers/azure/internal/compute.py
```diff
@@ -432,16 +436,20 @@ def GetOrCreateAnalysisVm(
...
+    logger.debug('VM image_reference {}'.format(image_reference))
```
giovannt0, line 445:
```suggestion
    logger.debug('VM image_reference {0:!s}'.format(image_reference))
```

tools/az_cli.py
```diff
@@ -100,6 +101,10 @@ def StartAnalysisVm(args: 'argparse.Namespace') -> None:
       logger.error('error: parameter --attach_disks: {0:s}'.format(
           args.attach_disks))
       return
+  if args.image_reference:
+    image_reference = json.loads(args.image_reference)
```
giovannt0, line 105: This can throw a `JSONDecodeError` or a `TypeError`, and should be wrapped accordingly

giovannt0 (changes requested):

libcloudforensics/providers/azure/internal/account.py
juju4: Done

tools/az_cli.py
juju4: Done

juju4:
Changes done.
I'm fine with CLA but I won't change github settings for it.
If it helps, consider contribution as public domain and merged separately.

libcloudforensics/providers/azure/internal/account.py
```diff
@@ -42,7 +42,7 @@ class AZAccount:
   """
 
   def __init__(self,
-               default_resource_group_name: str,
+               default_resource_group_name: str = '',
```
giovannt0, line 45: Can you update the docstring to reflect that this parameter is now Optional?

giovannt0 (changes requested):

libcloudforensics/providers/azure/internal/account.py
juju4: done

juju4:
any status/blocker on this PR?
change requested seems to be done

ramo-j:
Hi @juju4 

I don't think we can merge this without the CLA. 

> If it helps, consider contribution as public domain and merged separately.

I'm no lawyer, but I don't think this would help, as it won't change the authorship of the code - you would still be the author, even if one of us made an identical PR with our name on it.

tomchop:
@juju4 , I don't think accepting the CLA requires you to change any GitHub settings; all you need to do is go through the instructions here https://cla.developers.google.com/ and click through the prompts.

juju4:
Tried again for CLA but my reading is still that this require a change to github associated email(s)

"Your contribution (commit) must be associated with at least one of:
    The primary email on your Google Account associated with the signed CLA.
    An Alternate email on your Google Account associated with the signed CLA. See and modify your aternate emails at https://myaccount.google.com/alternateemail.
    The GitHub username associated with the signed CLA."

Only github username not matched to a google account (third case) does not seem to allow to sign it.
Always redirected to accounts.google.com

PR #431, opened 2022-02-18
Updated snapshot behaviour for when the resource exists already

ramo-j:
Currently implementation is that when attempting to snapshot a disk in GCE, if the snapshot exists already, an error is thrown. This PR captures that and returns the existing snapshot in such an event. The creation is also tracked, so the snapshot is only deleted if the snapshot was created anew. 

Also, unrelated updates for Azure tests.


tomchop (approved):
Small nit, otherwise LGTM!

PR #433, opened 2022-02-21
Added in retry on CreateDiskFromSnapshot when snapshot isn't yet ready

ramo-j:
Pretty much what it says on the tin - When the snapshot being used to create a disk isn't ready yet, wait for the snapshot to become ready.

### DIFF
--- a/libcloudforensics/providers/gcp/internal/compute.py
+++ b/libcloudforensics/providers/gcp/internal/compute.py
@@ -401,9 +401,17 @@ class GoogleCloudCompute(common.GoogleCloudComputeClient):
     }
     try:
       gce_disks_client = self.GceApi().disks()
-      request = gce_disks_client.insert(
-          project=self.project_id, zone=self.default_zone, body=body)
-      response = request.execute()
+      while True:
+        try:
+          request = gce_disks_client.insert(
+              project=self.project_id, zone=self.default_zone, body=body)
+          response = request.execute()
+          break
+        except HttpError as exception:
+          if not 'is not ready' in str(exception):
+            raise exception
+          logger.warning("Resource not yet ready, pausing 30 seconds...")
+          time.sleep(30)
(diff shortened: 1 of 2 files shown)


libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -401,9 +401,17 @@ def CreateDiskFromSnapshot(
...
-          project=self.project_id, zone=self.default_zone, body=body)
-      response = request.execute()
+      while True:
+        try:
+          request = gce_disks_client.insert(
+              project=self.project_id, zone=self.default_zone, body=body)
+          response = request.execute()
+          break
+        except HttpError as exception:
+          if not 'is not ready' in str(exception):
```
ramo-j, line 411:
```suggestion
          if 'is not ready' not in str(exception):
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -401,9 +401,17 @@ def CreateDiskFromSnapshot(
...
-          project=self.project_id, zone=self.default_zone, body=body)
-      response = request.execute()
+      while True:
+        try:
+          request = gce_disks_client.insert(
+              project=self.project_id, zone=self.default_zone, body=body)
+          response = request.execute()
+          break
+        except HttpError as exception:
+          if 'is not ready' not in str(exception):
```
jonathan-greig, line 411: Rather than retrying until it's ready is there somewhere that we can use [BlockOperation](https://github.com/google/cloud-forensics-utils/blob/985c723b5684b6c3caedb9d16ef67bf6abf34562/libcloudforensics/providers/gcp/internal/common.py#L240) to ensure this function isn't called until the resources are ready?

ramo-j:
As per OOB discussion - Closing this out for another tactic.

PR #434, opened 2022-02-22
Add randomness to snapshot names

ramo-j:
If there is parallel usage of the CopyDisk operation, the intermediate snapshots can end up with the same names and an error to create the snapshot, since it is based on timestamp. Adding some randomness to the name.

### DIFF
--- a/libcloudforensics/providers/gcp/internal/common.py
+++ b/libcloudforensics/providers/gcp/internal/common.py
@@ -152,6 +152,7 @@ def GenerateUniqueInstanceName(
     name = name[:truncate_at]
   return name
 
+
 def CreateService(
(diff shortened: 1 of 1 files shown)


libcloudforensics/providers/gcp/internal/common.py
```diff
@@ -143,10 +145,11 @@ def GenerateUniqueInstanceName(
     str: The name after adding a timestamp.
         Ex: [prefix]-[TIMESTAMP('%Y%m%d%H%M%S')]
```
jonathan-greig, line 146: Change comment to reflect new random component otherwise LGTM.

jonathan-greig (approved):

PR #440, opened 2022-03-29
Compute resources names are treated as project-wide unique identifiers

### LINKED ISSUE #438: Add support for specifying resources by ID as well as name
Add support for accessing resources by ID as well as name, for example instance ID for GetInstance.

### LINKED ISSUE #439: Instance names are treated as project-wide unique identifiers
## Issue Description
Instance names are only unique within a zone, not project wide.

For example [libcloudforensics.providers.gcp.internal.compute.ListInstances](https://github.com/google/cloud-forensics-utils/blob/9a0a583c12e3d17d677d396f985bdbe019c1ff26/libcloudforensics/providers/gcp/internal/compute.py#L124) generates a dictionary of all instances in a project keyed on instance name. If there are more than one instance in the project with the same name only one will be included in the returned dict.

## Possible Fixes:
1. Use instance IDs instead of instance names - a less human-friendly value but helper functions could be added to still access instances by name etc.
2. Use zone:instance_name to refer to instances.

jonathan-greig:
Fixes #439 

Changes `libcloudforensics.providers.gcp.internal.ListInstances` to return a dictionary keyed on instance ID instead of instance name which isn't unique project wide (#439).


libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -299,39 +360,50 @@ def ListRegionDisks(self) -> Dict[str, 'GoogleRegionComputeDisk']:
...
+      self,
+      disk_name: str,
+      region: Optional[str] = None) -> 'GoogleRegionComputeDisk':
     """Get regional disk in project.
 
     Regional disks API resource: https://cloud.google.com/compute/docs/reference/rest/v1/regionDisks#resource:-disk  # pylint: disable=line-too-long
 
     Args:
-      disk_name: name of the regional disk to get.
+      disk_name: Name of the regional disk to get.
```
sa3eed3ed, line 377:
```suggestion
      disk_name: Name or ID of the regional disk to get.
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -252,16 +313,16 @@ def ListDisks(self) -> Dict[str, 'GoogleComputeDisk']:
...
-        try:
-          for disk in response['items'][zone]['disks']:
-            _, zone = disk['zone'].rsplit('/', 1)
-            name = disk['name']
-            disks[name] = GoogleComputeDisk(
-                self.project_id, zone, name, labels=disk.get('labels'))
-        except KeyError:
-          pass
-
+      for location in response['items']:
```
sa3eed3ed, line 316:
```suggestion
      # Disks.aggregatedList returns both zones and regions in items.
      for location in response['items']:
```
```suggestion
      for location in response['items']:
```

libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -252,16 +313,16 @@ def ListDisks(self) -> Dict[str, 'GoogleComputeDisk']:
...
+        for disk in response['items'][location].get('disks', []):
+          # Skip if regional disk, i.e. has no region.
```
sa3eed3ed, line 318:
```suggestion
          # Skip if regional disk, i.e. has no zone.
```

PR #441, opened 2022-03-29
Fix new pylint errors in main

jonathan-greig:
* Added `# pylint: disable=no-member` on no-member errors (GCP libs dynamically add attributes so can't be statically checked)
* Fixed occurrences of `consider-using-dict-items`


ramo-j (approved):
LGTM

PR #447, opened 2022-06-08
Bugfix: truncate name before regex validation

tomchop:
The regex would never be valid if the name was too long. Conversely, the truncate statement would never have any effect, as only short-enough names would get truncated.


tomchop:
Failing tests seem to be related to Azure, and not this change.

giovannt0 (approved):

PR #456, opened 2022-08-29
Broken Azure unit test

### LINKED ISSUE #455: Broken Azure unit test
Patch in `cloud-forensics-utils/tests/providers/azure/internal/test_storage.py` needs to be updated for new version of Azure libs.

jonathan-greig:
Fixes #455


ramo-j:
This will keep happening every time there's a new version released. Is there a way to extract the version that gets installed, and then dynamically mock it?

Something like the following comes to mind
```python
version = extractVersion()
@mock.patch(f'azure.mgmt.storage.{version}.operations._storage_accounts_operations.StorageAccountsOperations.list_keys')
```

ramo-j (approved):

PR #481, opened 2023-06-01
Fix deleteInstance when no VPC

### LINKED ISSUE #462: instance.Delete() fails if no VPC is attached
instance.Delete() fails if no VPC is attached.

Failing example code snippet:
```python
instance = account.ec2.GetInstanceById(instance_id)
instance.Delete()
```

A key error  is thrown at:
https://github.com/google/cloud-forensics-utils/blob/a71b13c3a7108e4d37879007617203c5e34170ff/libcloudforensics/providers/aws/internal/ec2.py#L250

Fryyyyy:
Allow for a None value of VPC, which isn't used in the Delete call anyway.
No new EC2 instances are allowed to be created without a VPC by AWS so this only applies to old ones.
Fixes #462


ramo-j:
> No new EC2 instances are allowed to be created without a VPC by AWS so this only applies to old ones.

I think very recently, AWS turned off EC2 classic, so maybe this isn't even a thing any more.

ramo-j (approved):

Issue #487, opened 2023-09-07
Add support for regional GKE Clusters


dfjxs:
`GKEEnumerate` and `GKEWorkloadQuarantine` currently only supports zonal clusters.

GKE Autopilot [only allows regional clusters](https://cloud.google.com/kubernetes-engine/docs/concepts/types-of-clusters#config-choices). Support should be added for regional clusters.

PR #492, opened 2023-09-26
Add support for regional GKE Clusters

dfjxs:
Adding support for regional GKE Clusters to the `GKEEnumerate` and `GKEWorkloadQuarantine` methods.

Fixes: #487


libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -242,30 +242,56 @@ def ListMIGSByInstanceName(self, zone: str) -> Dict[str, str]:
...
     """
     groups_client = self.GceApi().instanceGroupManagers() # pylint: disable=no-member
-    responses = common.ExecuteRequest(
-        groups_client, 'list', {
-            'project': self.project_id,
-            'zone': zone,
-        })
-
     groups = defaultdict(list)
+    try:
```
sa3eed3ed, line 260: instead of doing this, maybe you can use a regex to determine if the location is a zone, I did similar before, this would work well:
```
ZONE_REGEX = re.compile(
    r'\A(?P<zone>(?P<region>(?P<geo_tag>[a-zA-Z0-9]+)-[a-zA-Z0-9]+)'
    r'-[a-zA-Z0-9]+)\Z'
)
```

sa3eed3ed (commented):
small drive-by observation on detecting if location is zone or region, thanks!

libcloudforensics/providers/gcp/internal/compute.py
dfjxs: Updated, thanks.

libcloudforensics/providers/gcp/forensics.py
```diff
@@ -583,7 +583,7 @@ def FirewallNodes() -> None:
     """Puts each node from the workload into network quarantine."""
     for node in workload_nodes:
       logger.info(
-          'Putting instance {0:s} into network quarantine...'.format(
+          'Putting workload {0:s} into network quarantine...'.format(
               workload_id))
```
sa3eed3ed, line 587:
```suggestion
          ('Putting instance "{0:s}" from workload "{1:s}" into network'
           ' quarantine...').format(node.name, workload_id))
```

sa3eed3ed (approved):

PR #495, opened 2023-11-06
Outbound network traffic

Fryyyyy:
Add a metrics capability to graph outbound network flow through the Metrics API

### DIFF
--- a/libcloudforensics/providers/gcp/internal/monitoring.py
+++ b/libcloudforensics/providers/gcp/internal/monitoring.py
@@ -106,16 +106,16 @@ class GoogleCloudMonitoring:
       days (int): Optional. The number of days to collect metrics for.
 
     Returns:
-      List[Dict[str, Any]]: a list of network usage for each instance in the format
+      List[Dict[str, Any]]: a list of network usage for each instance
+      in the format
         [
           {
             'instance_name': str,
-            'instance_id': str,
-            'cpu_usage':
+            'network_usage':
             [
               {
                 'timestamp': str,
-                'cpu_usage': float
+                'bytes': float
               },
             ]
           },
--- a/libcloudforensics/providers/gcp/internal/monitoring.py
+++ b/libcloudforensics/providers/gcp/internal/monitoring.py
@@ -91,6 +91,90 @@ class GoogleCloudMonitoring:
               ret[service] = int(val)
     return ret
 
+
+  def GetNetworkData(
+      self, instance_ids: Optional[List[str]] = None,
+      days: int = 2) -> List[Dict[str, Any]]:
+    """Returns outbound network metrics for compute instances.
+
+    By default returns usage for the last two days for the given instance.
+
+    Args:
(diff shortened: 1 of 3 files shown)


libcloudforensics/providers/gcp/internal/monitoring.py
```diff
@@ -91,6 +91,90 @@ def ActiveServices(self, timeframe: int = 30) -> Dict[str, int]:
               ret[service] = int(val)
     return ret
 
+
+  def GetNetworkData(
+      self, instance_ids: Optional[List[str]] = None,
```
ramo-j, line 96:
```suggestion
       self, instance_ids: List[str] = [],
```
Style nit: Removing Optional, and having a default of a zereo length list might be preferable here?

libcloudforensics/providers/gcp/internal/monitoring.py
```diff
@@ -91,6 +91,90 @@ def ActiveServices(self, timeframe: int = 30) -> Dict[str, int]:
...
+      days: int = 2) -> List[Dict[str, Any]]:
```
ramo-j, line 97: Is it possible that a user might want to specify a start and end date? Rather than just "the last N days."

libcloudforensics/providers/gcp/internal/monitoring.py
```diff
@@ -91,6 +91,90 @@ def ActiveServices(self, timeframe: int = 30) -> Dict[str, int]:
...
+        'resource.type="gce_instance"'
+    ]
+
+    if instance_ids:
+      network_filter.append(
+          ' AND (resource.label.instance_id = "{0:s}"'.format(instance_ids[0]))
+      if len(instance_ids) > 1:
+        for instance_name in instance_ids[1:]:
+          network_filter.append(
+              ' OR resource.label.instance_id = "{0:s}"'.format(instance_name))
```
ramo-j, line 142: Based on [0] it looks like you could simplify this as:
```suggestion
      ids = ' OR '.join(f'"id"' for id in instance_ids)
      network_filter.append(f' AND (resource.label.instance_id = ({ids}))')
```

[0] https://cloud.google.com/logging/docs/view/logging-query-language

ramo-j (approved):
A few notes, otherwise LGTM

libcloudforensics/providers/gcp/internal/monitoring.py
Fryyyyy: Probably correct but will leave it to match the existing code patterns for now

libcloudforensics/providers/gcp/internal/monitoring.py
Fryyyyy: Its possible but will leave it to match the existing code patterns for now

libcloudforensics/providers/gcp/internal/monitoring.py
Fryyyyy: Probably correct but will leave it to match the existing code patterns for now

PR #502, opened 2024-02-14
Bump urllibl3 version to 2.0.7 and fix Azure unit tests

jleaniz:
Bump urllib3 version requirements to match other dependencies like boto3/botocore which support 2.0.7. This also would unblock Turbinia from pinning to urllib3 1.26 which is on its way to being deprecated


tomchop (approved):

tomchop:
LGTM, but we should fix everything else that's failing?

PR #530, opened 2025-03-31
Force images of hyperdisks to be regular disks

Fryyyyy:
Fixes:
```
HttpError 400 when requesting https://compute.googleapis.com/com
pute/v1/projects/project-name/prod/zones/us-central1-f/instances/test-vm/attachDisk?alt=json returned "hyperdisk-balanced disk type cannot be attached in read-only mode.". Details
: "[{'message': 'hyperdisk-balanced disk type cannot be attached in read-only mode.', 'domain': 'global', 'reason': 'badRequest', 
```


Fryyyyy:
Pylint complaints are annoying and not relevant

ramo-j (approved):

PR #532, opened 2025-07-21
Add support for image formats when exporting to GCS

ramo-j:
Also: 

* command line / reference example to invoke the exporting of a disk image to GCS.
* Extensive linter corrections


Fryyyyy:
Tests plz

Fryyyyy (approved):

PR #541, opened 2026-04-14
Allow zones to be provided for disks and instances

Fryyyyy:
Rather than listing all of them, which causes timeouts in large projects


libcloudforensics/providers/gcp/internal/compute.py
```diff
@@ -143,6 +143,73 @@ def _FindResourceByName(
...
+      zone: Optional zone to restrict the search.
+      region: Optional region to restrict the search.
+
+    Returns:
+      Optional[Dict[str, Any]]: The resource metadata if found, None otherwise.
+    """
+    client = getattr(self.GceApi(), resource_type)()
+    if zone or region:
+      param_name = resource_type[:-1] if resource_type.endswith(
+          's') else resource_type
```
ramo-j: Nit: This reads kind of awkward, how about:
```suggestion
      param_name = (resource_type[:-1] if resource_type.endswith('s')
                    else resource_type)
```

ramo-j (approved):
Just one readability nit, otherwise, LGTM.

